PR #52, opened 2015-11-03
Add logging for Kong and Dynamo

marialivia16:


app/controllers/Application.scala
```diff
@@ -69,15 +69,21 @@ class Application(dynamo: DB, kong: Kong, val messagesApi: MessagesApi, val auth
...
-
-    Ok(views.html.editUser(id, filledForm, request.user.firstName, userKeys, editUserPageTitle))
+    dynamo.getUserWithId(id) match {
+      case Some(consumer) => {
+        val filledForm = editUserForm.fill(EditUserFormData(consumer.name, consumer.email, consumer.productName, consumer.productUrl, consumer.companyName, consumer.companyUrl))
+        val userKeys = dynamo.getKeysWithUserId(id)
+        Ok(views.html.editUser(id, filledForm, request.user.firstName, userKeys, editUserPageTitle))
+      }
+      case None => {
+        val userKeys = dynamo.getKeysWithUserId(id)
```
oilnam: We might want to move `val userKeys = dynamo.getKeysWithUserId(id)` outside the match since we use it twice?

marialivia16: @oilnam Yes, thanks! I didn't see that.

app/store/Dynamo.scala
```diff
@@ -10,88 +10,178 @@ import play.api.Logger
...
+  }
+
+  def getUserWithId(id: String): Option[BonoboUser] = {
     val userQuery = new QuerySpec()
       .withKeyConditionExpression("id = :i")
       .withValueMap(new ValueMap().withString(":i", id))
       .withMaxResultSize(1)
-    BonoboTable.query(userQuery).asScala.toList.map(fromBonoboItem).head
+    val resultUser = BonoboTable.query(userQuery).asScala.toList.map(fromBonoboItem).headOption
+    Logger.info(s"DynamoDB: User with id $id is $resultUser")
```
oilnam: Isn't logging "get" requests to the db a bit too much?

marialivia16: @oilnam It is, I thought about it. Sometimes they are useful, but most of the time they are not. I think we can remove that log.

oilnam: I'd remove all the "read" logs and only keep those that alter the state of the database (write/update).

app/controllers/Application.scala
```diff
@@ -69,15 +69,20 @@ class Application(dynamo: DB, kong: Kong, val messagesApi: MessagesApi, val auth
   }
 
   def editUserPage(id: String) = maybeAuth { implicit request =>
-    val consumer = dynamo.getUserWithId(id)
-    val userKeys = dynamo.getAllKeysWithId(id)
-    val filledForm = editUserForm.fill(EditUserFormData(consumer.name, consumer.email, consumer.productName, consumer.productUrl, consumer.companyName, consumer.companyUrl))
-
-    Ok(views.html.editUser(id, filledForm, request.user.firstName, userKeys, editUserPageTitle))
+    val userKeys = dynamo.getKeysWithUserId(id)
+    dynamo.getUserWithId(id) match {
```
cb372: Ideally this kind of fix should be in a separate PR, as it has nothing to do with logging. 

It's fine to keep it in the PR this time, but it's good to get into the habit of keeping unrelated changes in separate PRs. It makes them easier to review, and it means that we can rollback one change (if it turns out to be buggy) without losing the other.

oilnam:
:+1:

PR #54, opened 2015-11-04
Add search by product name and keep query parameter in search box

marialivia16:
- Fix search by company name
- Add search by product name
- While on search page show the query parameter in the search box so that it can be easily changed


app/views/main.scala.html
```diff
@@ -17,7 +17,7 @@
                 <div id="navbar" class="navbar-collapse collapse">
                     <form action="@routes.Application.search" class="navbar-form navbar-left" role="search" method="GET">
                         <div class="form-group">
-                            <input type="text" class="form-control" placeholder="Search" name="query">
+                            <input type="text" class="form-control" placeholder="Search" name="query" value="@{if(query.isDefined) query.get}">
```
cb372: I haven't tried it, but I'm pretty sure Play templates have special support for `Option`, so you can just write `value="@query"` and it will do the right thing.

cb372:
:+1: apart from one comment

PR #55, opened 2015-11-04
OpenForm refactor and cleanup

oilnam:
I modified `createUser` to return a key instead of a userId. This has the following advantages:

* code is shorter and saves a db query
* `dynamo.getKeyWithUserId` becomes useless (we already had `getKeysWithUserId` btw)
* the URL after a user has been created through OpenForm is now `?key=<key>` instead of `?userId=<userId>`, which is slightly better since we don't disclose internal information.

### DIFF
--- a/app/controllers/OpenForm.scala
+++ b/app/controllers/OpenForm.scala
@@ -15,7 +15,7 @@ class OpenForm(dynamo: DB, kong: Kong, val messagesApi: MessagesApi) extends Con
   import OpenForm._
   import Forms.OpenCreateKeyFormData
 
-  private val openLogic = new OpenFormLogic(dynamo, kong)
+  private val logic = new OpenFormLogic(dynamo, kong)
 
   def createKeyPage = Action { implicit request =>
     Ok(views.html.openCreateKey(createKeyForm))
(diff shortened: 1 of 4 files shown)


app/logic/OpenFormLogic.scala
```diff
@@ -20,6 +20,7 @@ class OpenFormLogic(dynamo: DB, kong: Kong) {
    *
    * @return a Future of the newly created Kong consumer's ID
```
cb372: This comment needs to be updated to match the new behaviour of the method (now it returns the key value, not the consumer ID).

cb372:
Good stuff. :+1: apart from one comment.

app/controllers/OpenForm.scala
```diff
@@ -15,7 +15,7 @@ class OpenForm(dynamo: DB, kong: Kong, val messagesApi: MessagesApi) extends Con
   import OpenForm._
   import Forms.OpenCreateKeyFormData
 
-  private val logic = new OpenFormLogic(dynamo, kong)
+  private val openLogic = new OpenFormLogic(dynamo, kong)
```
marialivia16: I do not understand why we need this renaming. We are not using the variable outside of this class, it is clear what it does. Plus, we are not being consistent with the other `logic` variable.

oilnam: You're right (:

PR #58, opened 2015-11-05
Improve search, remove duplicates and sort results

marialivia16:
We were scanning DynamoDb for equality `email = :some_value`, but in the case of a name (for example) we might want to search only by the last name. Using the`contains(email, :some_value)` returns more relevant results in most of the cases.

### DIFF
--- a/app/store/Dynamo.scala
+++ b/app/store/Dynamo.scala
@@ -41,7 +41,7 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String) extends DB {
 
   def search(query: String, limit: Int = 20): List[BonoboInfo] = {
     val keysScan = new ScanSpec()
-      .withFilterExpression("#1 = :s")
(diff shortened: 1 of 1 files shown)


cb372:
:+1: Seems a little weird that we have to sort the list again even though Dynamo returned sorted results to us, but I understand the reason.

It's a shame it's not possible to do case-insensitive queries in Dynamo. We could fix this in the future by adding extra fields to Dynamo to store lowercase versions of the fields, but it's low priority.

Unrelated to this PR, I noticed that we are doing N+1 Dynamo queries for both the search page and the list keys page. We get a list of (up to 20) keys, then for each one we make a separate query to Dynamo to get the associated user. We could do this more efficiently by collecting all the user IDs we need and then doing just one query to find all the users.

marialivia16:
@cb372 Too bad there isn't a better workaround to do case-insensitive queries. 
In a separate PR I'll make the queries more effective.

PR #59, opened 2015-11-06
Redesign open registration form

marialivia16:
Update open registration form to render fields on 2 columns.


LATaylor-guardian:
:+1:

PR #60, opened 2015-11-09
Add commercial registration form

marialivia16:
What it does: The form saves the information the user has provided in DynamoDb. New fields have been added to the model (additional info) and the edit user page is now displaying more data.

What it doesn't do: Send email to notify administrators that a new request has been submitted and to notify the user that a key has been created. This will be done in a separate PR.

[Form design](https://github.com/guardian/bonobo/files/29872/Screen.Shot.2015-11-09.at.10.15.50.pdf)


app/components/AppComponents.scala
```diff
@@ -67,8 +66,9 @@ trait ControllersComponent { self: BuiltInComponents with NingWSComponents with
   def authController = new Auth(googleAuthConfig, wsApi)
 
   val openFormController = new OpenForm(dynamo, kong, messagesApi)
```
cb372: Maybe we should rename the "open form" to "developer form" for consistency?

app/controllers/Application.scala
```diff
@@ -73,25 +73,28 @@ class Application(dynamo: DB, kong: Kong, val messagesApi: MessagesApi, val auth
     dynamo.getUserWithId(id) match {
       case Some(consumer) => {
         val filledForm = editUserForm.fill(EditUserFormData(consumer.name, consumer.email, consumer.productName, consumer.productUrl, consumer.companyName, consumer.companyUrl))
-        Ok(views.html.editUser(id, filledForm, request.user.firstName, userKeys, editUserPageTitle))
+        Ok(views.html.editUser(id, filledForm, Some(consumer.additionalInfo), request.user.firstName, userKeys, editUserPageTitle))
       }
       case None => {
-        BadRequest(views.html.editUser(id, editUserForm, request.user.firstName, userKeys, editUserPageTitle, error = Some("Something bad happened while getting the user from the database.")))
+        BadRequest(views.html.editUser(id, editUserForm, None, request.user.firstName, userKeys, editUserPageTitle, error = Some("Something bad happened while getting the user from the database.")))
```
cb372: This should actually be a `NotFound` (i.e. a 404), not a `BadRequest`. And the error should say something like "User not found".

app/models/models.scala
```diff
@@ -12,20 +12,37 @@ case class BonoboUser(
...
 
 object BonoboUser {
   def apply(id: String, formData: CreateUserFormData): BonoboUser = {
-    new BonoboUser(id, formData.email, formData.name, formData.productName, formData.productUrl, formData.companyName, formData.companyUrl)
+    val additionalInfo = AdditionalUserInfo(DateTime.now(), ManualRegistration, None, None, None, None, None)
+    new BonoboUser(id, formData.email, formData.name, formData.productName, formData.productUrl, formData.companyName, formData.companyUrl, additionalInfo)
   }
-  def apply(id: String, formData: EditUserFormData): BonoboUser = {
-    new BonoboUser(id, formData.email, formData.name, formData.productName, formData.productUrl, formData.companyName, formData.companyUrl)
+  def apply(id: String, formData: EditUserFormData, createdAt: DateTime, registrationType: RegistrationType): BonoboUser = {
```
cb372: It's a little confusing having 4 `apply`methods. I would add some comments to each one to explain its use case, e.g. this one is used when manually creating a user, ...

app/models/models.scala
```diff
@@ -12,20 +12,37 @@ case class BonoboUser(
...
+    new BonoboUser(id, formData.email, formData.name, formData.productName, formData.productUrl, formData.companyName, formData.companyUrl, additionalInfo)
+  }
+  def apply(formData: CommercialRequestKeyFormData): BonoboUser = {
+    val additionalInfo = AdditionalUserInfo(DateTime.now(), CommercialRegistration, formData.businessArea, formData.monthlyUsers.map(_.toString()), formData.commercialModel, formData.content, formData.articlesPerDay.map(_.toString()))
+    new BonoboUser(java.util.UUID.randomUUID().toString, formData.email, formData.name, formData.productName, formData.productUrl, formData.companyName, formData.companyUrl, additionalInfo)
   }
 }
 
+case class AdditionalUserInfo(
+  createdAt: DateTime,
```
cb372: Shouldn't `createdAt` and `registrationType` be on the main `BonoboUser` class? Because they are defined for all users, whereas `businessArea`, etc. are only relevant for commercial users.

conf/routes
```diff
@@ -17,12 +17,18 @@ POST        /user/create               controllers.Application.createUser
...
 GET         /healthcheck               controllers.Application.healthcheck
 
+#Developers open registration form
+GET         /formDeveloper             controllers.OpenForm.createKeyPage
+POST        /formDeveloper             controllers.OpenForm.createKey
+GET         /developerKey              controllers.OpenForm.showKey(key: String)
+
+#Commercial request key form
+GET         /formCommercial            controllers.CommercialForm.requestKeyPage
+POST        /formCommercial            controllers.CommercialForm.requestKey
```
cb372: Sorry, I have an irrational hatred of camel-case URLs. Can we change these to `/register/developer` and `/register/commercial`?

app/models/models.scala
marialivia16: @cb372 I was thinking about it yesterday and I ended up storing in `AdditionalUserInfo` the data I need on the page, but it is not in the form. If I remove `createdAt` and `registrationType` I will have to send 2 extra parameters to the template. Do you think this is a better approach?

PR #63, opened 2015-11-12
Set page size to 20

cb372:


app/store/Dynamo.scala
```diff
@@ -24,13 +24,13 @@ trait DB {
...
 
-  def getKeys(direction: String, range: Option[String]): ResultsPage[BonoboInfo]
+  def getKeys(direction: String, range: Option[String], limit: Int = 20): ResultsPage[BonoboInfo]
 
   def getKeyWithValue(key: String): Option[KongKey]
 
   def getKeysWithUserId(id: String): List[KongKey]
 
-  def getNumberOfKeys: Long
+  def getNumberOfKeys(): Long
```
LATaylor-guardian: Why have you added ()?

LATaylor-guardian:
:+1:

app/store/Dynamo.scala
cb372: Because it has a side effect, namely a DynamoDB API call. By convention, anything with a side effect should have parentheses.

PR #64, opened 2015-11-12
Migrate endpoint

marialivia16:
- Migration endpoint has been added in `controllers/Migration.scala`. The endpoint accepts a POST with a json body containing a list of users and saves a new consumer with his keys to Kong and DynamoDb. If the key is marked as `Inactive` it will be deleted from Kong.
- Use enumeratum for Tier model.
- Tests and a json example file have been added.
- Returns a Json containing:
    - the number of successfully added users
   - the number of successfully added keys
   - the list of users with conflicts
   - the list of keys with conflicts

### DIFF
--- a/app/controllers/Migration.scala
+++ b/app/controllers/Migration.scala
@@ -36,7 +36,7 @@ class Migration(dynamo: DB, kong: Kong) extends Controller {
   private def handleUserAndKeys(bonoboUser: BonoboUser, masheryKeys: List[MasheryKey]): Future[List[Unit]] = {
     dynamo.getUserWithEmail(bonoboUser.email) match {
       case Some(user) => {
-        Logger.warn(s"Migration: Email already taken when creating user with name ${bonoboUser.name} and keys with value ${masheryKeys.head.key}")
+        Logger.warn(s"Migration: Email already taken when creating user with name ${bonoboUser.name} and keys ${masheryKeys.map(_.key)}")
         Future.failed(ConflictFailure("Email already taken. You cannot have two users with the same email."))
       }
       case None => {
--- a/app/controllers/Migration.scala
+++ b/app/controllers/Migration.scala
@@ -1,10 +1,9 @@
 package controllers
 
 import kong.Kong
-import kong.Kong.ConflictFailure
 import models._
 import play.api.Logger
-import play.api.libs.json.{ JsError, JsSuccess }
+import play.api.libs.json.{ Json, JsError, JsSuccess }
 import play.api.mvc._
 import store.DB
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -15,7 +14,11 @@ class Migration(dynamo: DB, kong: Kong) extends Controller {
   def migrate = Action.async(parse.json) { implicit request =>
     request.body.validate[List[MasheryUser]] match {
       case JsSuccess(masheryUsers, _) => {
-        Future.traverse(masheryUsers)(masheryUser => handleMasheryUser(masheryUser)) map (_ => Ok("OK"))
+        Future.traverse(masheryUsers)(masheryUser => handleMasheryUser(masheryUser)) map { userResults =>
+          val count = handleUserResult(userResults)
+          Logger.info(s"Number of successful users: $count")
+          Ok(Json.toJson(count))
+        }
       }
       case JsError(errorMessage) => {
         Logger.warn(s"Migration: Error when parsing json: $errorMessage")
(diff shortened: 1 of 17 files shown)


cb372:
I'll review this after the changes in #65 have been applied and it merges cleanly.

app/models/models.scala
```diff
@@ -146,4 +149,34 @@ case object CommercialRegistration extends RegistrationType {
...
+case class MasheryUser(
+  name: String,
+  email: String,
+  companyName: String,
+  companyUrl: String,
+  createdAt: DateTime,
+  keys: List[MasheryKey])
+
+object MasheryKey {
+  implicit val dateReads = Reads.jodaDateReads("MMM dd yyyy HH:mm:ss")
```
cb372: Could we use standard ISO date time formats for the dates please? It makes life slightly easier for my migration script, and using standard date formats is a good idea in general.

app/controllers/Migration.scala
```diff
@@ -0,0 +1,70 @@
...
+    val bonoboId = java.util.UUID.randomUUID().toString
+    val additionalInfo = AdditionalUserInfo(user.createdAt, MasheryRegistration)
+    val bonoboUser = BonoboUser(bonoboId, user.email, user.name, unspecifiedIfEmpty(user.companyName), unspecifiedIfEmpty(user.companyUrl), additionalInfo)
+    handleUserAndKeys(bonoboUser, user.keys)
+  }
+
+  private def handleUserAndKeys(bonoboUser: BonoboUser, masheryKeys: List[MasheryKey]): Future[List[Unit]] = {
+    dynamo.getUserWithEmail(bonoboUser.email) match {
+      case Some(user) => {
+        Logger.warn(s"Migration: Email already taken when creating user with name ${bonoboUser.name} and keys with value ${masheryKeys.head.key}")
```
cb372: There's a (very small) chance that the `masheryKeys` list will be empty (we might want to import a Mashery user who has no keys). So I would put `masheryKeys.map(_.key)` in the error message instead of `masheryKeys.head.key`

app/controllers/Migration.scala
```diff
@@ -0,0 +1,70 @@
...
+          {
+            val kongKey = KongKey(bonoboUser.bonoboId, consumer, RateLimits(masheryKey.requestsPerMinute, masheryKey.requestsPerDay), masheryKey.tier, masheryKey.productName, masheryKey.productUrl, masheryKey.status, masheryKey.createdAt)
+            dynamo.saveKey(kongKey)
+            if (masheryKey.status == KongKey.Inactive) kong.deleteKey(consumer.id).map(_ => ())
+            else Future.successful(())
+          }
+      }
+    }
+  }
+}
```
cb372: I don't think we should use `Future.failed` for the migration endpoint, because we will lose useful information. We'll get an error message saying e.g. "Email already taken ..." or "Key ... already taken", but we won't know anything about what happened to the other users and keys. How many were successful? Which other ones failed?

I would model the various kinds of success/failure results as sealed traits, and return detailed information about the result in the HTTP response. (This implies we always return a 200, even if there were problems.) That way we can decide whether to manually fix the problem keys, or re-run the migration, etc.

Rough sketch:

```scala
case class MigrationResult(successfullyMigratedUsers: Int, successfullyMigratedKeys: Int, userConflicts: List[UserConflict], keyConflicts: List[KeyConflict])

sealed trait MigrateUserResult
case class MigratedUser(keyResults: List[MigrateKeyResult]) extends MigrateUserResult
case class EmailConflict(email: String) extends MigrateUserResult

sealed trait MigrateKeyResult
case object MigratedKey extends MigrateKeyResult
case class KeyConflict(key: String) extends MigrateKeyResult

def migrate = Action.async(parse.json) { implicit request =>
  // ...
  Future.traverse(masheryUsers)(masheryUser => handleMasheryUser(masheryUser)) map { userResults =>
    // convert the results to a MigrationResult
  }
}

private def handleMasheryUser(user: MasheryUser): Future[MigrateUserResult] = {
  // ...
}

private def handleUserAndKeys(bonoboUser: BonoboUser, masheryKeys: List[MasheryKey]): Future[MigrateUserResult] = {
  dynamo.getUserWithEmail(bonoboUser.email) match {
    case Some(user) => Future.successful(EmailConflict(email))
    case None => 
      dynamo.saveUser(bonoboUser)
      Future.traverse(masheryKeys)(key => createKey(bonoboUser, key)).map { keyResults => MigratedUser(keyResults) }
}

private def createKey(bonoboUser: BonoboUser, masheryKey: MasheryKey): Future[MigrateKeyResult] = {
  if (dynamo.getKeyWithValue(masheryKey.key).isDefined)
    Future.successful(KeyConflict(masheryKey.key))
  else {
    // ...
    MigratedKey
  }    
}
```

app/models/models.scala
```diff
@@ -80,6 +82,10 @@ object KongKey {
     new KongKey(bonoboId, consumer.id, consumer.key, rateLimits.requestsPerDay, rateLimits.requestsPerMinute, tier, Active, consumer.createdAt, productName, productUrl, uniqueRangeKey(consumer.createdAt))
   }
 
+  /* model used in the migration endpoint */
+  def apply(bonoboId: String, consumer: ConsumerCreationResult, rateLimits: RateLimits, tier: Tier, productName: String, productUrl: String, status: String, date: DateTime): KongKey = {
+    new KongKey(bonoboId, consumer.id, consumer.key, rateLimits.requestsPerDay, rateLimits.requestsPerMinute, tier, status, date, productName, productUrl, uniqueRangeKey(consumer.createdAt))
```
cb372: @marialivia16 I noticed a bug here when I ran the migration script in CODE. The `uniqueRangeKey` should be created using the `date` that is passed into the method, not `consumer.createdAt`.

`consumer.createdAt` is the current time, so migrated keys end up being ordered by the time they were migrated into Bonobo, not the time they were created in Mashery.

marialivia16: @cb372 Ok, thank you for pointing that out. Should I restore this PR and make the changes here?

PR #65, opened 2015-11-17
Change model

marialivia16:
The `productName` and `productUrl` fields have been moved to the `KongKey` class.


cb372:
:+1: It definitely makes more sense to store these fields on a per-key level.

So with this change we can migrate from the Mashery model cleanly?

cb372:
:+1:

PR #66, opened 2015-11-20
Add field contentFormat and make fields required

marialivia16:
- Add the `contentFormat` field for the commercial form
- Make all form fields required (except for `key`)


cb372:
:+1: once the tests pass

PR #71, opened 2015-11-24
Add index on the users table email field in Dynamo

cb372:
and a method that uses that index to check if a given email address is already registered.

Doing this with a scan instead of an index query is unusably slow.


LATaylor-guardian:
This looks good to me, and some nice tidy up to make the code more readable. :+1:

PR #72, opened 2015-11-25
Handle empty fields

marialivia16:
Fields `name`, `productName` and `productUrl` are marked as `unspecified` if empty.

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -1,7 +1,7 @@
 # This is the main configuration file for the application.
 # ~~~~~
 kong.apiName = "internal"
-kong.apiAddress = "http://localhost:8001"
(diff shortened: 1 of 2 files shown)


conf/application.conf
```diff
@@ -1,7 +1,7 @@
 # This is the main configuration file for the application.
 # ~~~~~
 kong.apiName = "internal"
-kong.apiAddress = "http://localhost:8001"
+kong.apiAddress = "http://52.18.126.249:8001"
```
LATaylor-guardian: Did you mean to check this in?

marialivia16: No, I'm removing it now.

PR #73, opened 2015-11-25
Don't migrate things in parallel because Kong doesn't like it

cb372:
Kong occasionally generates duplicate UUIDs (!!) for consumer IDs. Work
around by only trying to create one key at a time.

Also better error handling and logging.

### DIFF
--- a/app/controllers/Migration.scala
+++ b/app/controllers/Migration.scala
@@ -8,59 +8,68 @@ import play.api.mvc._
 import store.DB
 import scala.concurrent.ExecutionContext.Implicits.global
 
-import scala.concurrent.Future
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
 
 class Migration(dynamo: DB, kong: Kong) extends Controller {
-  def migrate = Action.async(parse.json) { implicit request =>
+  def migrate = Action(parse.json) { implicit request =>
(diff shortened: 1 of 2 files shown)


app/controllers/Migration.scala
```diff
@@ -8,59 +8,68 @@ import play.api.mvc._
...
-      kong.createConsumerAndKey(masheryKey.tier, rateLimits, Option(masheryKey.key)) map {
-        consumer =>
-          {
-            val kongKey = KongKey(bonoboUser.bonoboId, consumer, RateLimits(masheryKey.requestsPerMinute, masheryKey.requestsPerDay), masheryKey.tier, unspecifiedIfEmpty(masheryKey.productName), unspecifiedIfEmpty(masheryKey.productUrl), masheryKey.status, masheryKey.createdAt)
-            dynamo.saveKey(kongKey)
-            if (masheryKey.status == KongKey.Inactive) kong.deleteKey(consumer.id)
-            MigratedKey
-          }
+  private def createKeyForUser(bonoboUser: BonoboUser, masheryKey: MasheryKey): MigrateKeyResult = {
+    try {
```
LATaylor-guardian: Wouldn't it be better to use a Scala 'Try' here?

LATaylor-guardian:
:+1: looks good.

PR #80, opened 2015-12-03
search button should be type submit

gidsg:
Previously when you click the search button nothing happened, because the type was not `submit` CC @LATaylor-guardian


LATaylor-guardian:
:+1:

PR #82, opened 2015-12-04
Add CSRF tokens

marialivia16:


cb372:
:+1:

LATaylor-guardian:
@mchv You like making PR's for Play don't you ;)

app/controllers/csrf/CSRFAction.scala
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  private[csrf] def isCached(result: Result): Boolean =
+    result.header.headers.get(CACHE_CONTROL).fold(false)(!_.contains("no-cache"))
+
+  private[csrf] def clearTokenIfInvalid(request: RequestHeader, config: CSRFConfig, errorHandler: ErrorHandler, msg: String): Future[Result] = {
+    import play.api.libs.iteratee.Execution.Implicits.trampoline
+
+    errorHandler.handle(request, msg) map { result =>
+      CSRFSupport.getToken(request).fold(
```
cb372: For the record, this is the line we needed to change, which resulted in us copy-pasting huge chunks of Play code.

The original code was calling `CSRF.getToken`, which relies on `CSRFConfig.global`, which relies on Play runtime DI, which we are not using. Result: MethodNotFoundError!

mchv: If not yet done, we should create an issue. I am happy to try to contribute a fix when I will have some time.

app/components/AppComponents.scala
```diff
@@ -74,6 +77,10 @@ trait AwsEmailComponentImpl extends AwsEmailComponent { self: BuiltInComponents
   }
 }
 
+trait CSRFComponent { self: BuiltInComponents =>
+  override lazy val httpFilters = Seq(CSRFFilter(CSRFConfig(), new ConfigTokenProvider(CSRFConfig())))
```
mchv: Why not using the existing `CSRFComponents` [code](https://github.com/playframework/playframework/blob/master/framework/src/play-filters-helpers/src/main/scala/play/filters/csrf/csrf.scala#L245-L254)?

app/controllers/csrf/CSRFAction.scala
mchv: Ok I think I know how to fix this, I will try to contribute a patch.

mchv: See https://github.com/playframework/playframework/pull/5347

PR #83, opened 2015-12-04
Send emails only if feature is enabled

marialivia16:
This will avoid sending emails in debug mode.

### DIFF
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -66,6 +66,7 @@ trait KongComponentImpl extends KongComponent { self: BuiltInComponents with Nin
 
 trait AwsEmailComponent {
   def awsEmail: MailClient
+
 }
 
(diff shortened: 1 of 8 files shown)


cb372:
ApplicationSpec doesn't compile :(

app/components/AppComponents.scala
```diff
@@ -105,4 +106,5 @@ class AppComponents(context: Context)
     with ControllersComponent
     with CSRFComponent {
   def enableAuth = true
+  def enableEmail: Boolean = configuration.getBoolean("email-enabled") getOrElse false
```
cb372: The key should be `email.enabled` to match the one in `application.conf`.

cb372:
As discussed offline, let's check the `emailEnabled` flag just once, when we create the email client, instead of passing the flag to the controllers.

cb372:
:+1: 

Please add `email.enabled = true` to the PROD config file on S3 before merging this.

PR #85, opened 2015-12-08
Add Open Platform header to commercial and developer forms

cb372:
This is because we can't include the forms in iframes on the
open-platform site, because we need to use cookies for CSRF.


LATaylor-guardian:
Might want to add this to the bottom of the pages (as is on the open platform site):

© 2015 Guardian News and Media Limited or its affiliated companies. All rights reserved.

PR #86, opened 2015-12-08
Make row clickable and remove edit link

marialivia16:
Instead of using the `Edit user`/`Edit key` links, the entire row is now clickable.


LATaylor-guardian:
:+1:

PR #88, opened 2015-12-15
Update README.md and add script for running locally

marialivia16:


scripts/configure-docker.sh
```diff
@@ -0,0 +1,43 @@
+#! /bin/bash
+
+container_host=${1?Input file parameter missing}
```
cb372: We could work out the container host from `$DOCKER_HOST`: 

```
docker_host=${DOCKER_HOST?Docker host is not set}
echo $docker_host | sed -e 's/tcp:\/\/\(.*\):.*/\1/'
```

PR #89, opened 2015-12-22
Label users

marialivia16:
- Add UI for adding labels (create and edit user pages)
- Save selected labels on Dynamo
- Edit labels associated with an user


app/components/AppComponents.scala
```diff
@@ -79,14 +81,27 @@ trait AwsEmailComponentImpl extends AwsEmailComponent { self: BuiltInComponents
...
 
+trait LabelsComponent {
+  def labelsMap: Map[String, LabelProperties]
+}
+
+trait LabelsComponentImpl extends LabelsComponent with DynamoComponent {
+  val labelsMap = {
+    val labelsList = dynamo.getLabels()
+    labelsList.foldLeft(Map.empty: Map[String, LabelProperties]) { (acc, label) =>
+      acc + (label.id -> LabelProperties(label.properties.name, label.properties.colour))
```
cb372: This could be simplified to `acc + (label.id -> label.properties)`

app/controllers/Application.scala
```diff
@@ -79,30 +79,37 @@ class Application(dynamo: DB, kong: Kong, awsEmail: MailClient, val messagesApi:
...
+        NotFound(views.html.editUser(id, editUserForm, None, None, labelsMap, request.user.firstName, userKeys, editUserPageTitle, error = Some("User not found.")))
     }
   }
 
   def editUser(id: String) = maybeAuth { implicit request =>
     val userKeys = dynamo.getKeysWithUserId(id)
-    val additionalInfo = dynamo.getUserWithId(id) match {
-      case Some(user) => Some(user.additionalInfo)
+    val user = dynamo.getUserWithId(id)
+    val additionalInfo = user match {
```
cb372:
```scala
val additionalInfo = user.map(_.additionalInfo)
val userLabels = user.map(_.labelIds)
```

app/controllers/Application.scala
```diff
@@ -79,30 +79,37 @@ class Application(dynamo: DB, kong: Kong, awsEmail: MailClient, val messagesApi:
...
     }
 
     def handleValidForm(form: EditUserFormData): Result = {
       logic.updateUser(id, form) match {
-        case Left(error) => Conflict(views.html.editUser(id, editUserForm.fill(form), additionalInfo, request.user.firstName, userKeys, editUserPageTitle, error = Some(error)))
-        case Right(_) => Ok(views.html.editUser(id, editUserForm.fill(form), additionalInfo, request.user.firstName, userKeys, editUserPageTitle, success = Some("The user has been successfully updated.")))
+        case Left(error) => Conflict(views.html.editUser(id, editUserForm.fill(form), additionalInfo, userLabels, labelsMap, request.user.firstName, userKeys, editUserPageTitle, error = Some(error)))
+        case Right(_) => Redirect(routes.Application.editUserPage(id)).flashing("success" -> "The user has been successfully updated.") //Ok(views.html.editUser(id, editUserForm.fill(form), additionalInfo, userLabels, labelsMap, request.user.firstName, userKeys, editUserPageTitle, success = Some("The user has been successfully updated.")))
```
cb372: Please delete the commented-out code

app/logic/ApplicationLogic.scala
```diff
@@ -36,7 +36,12 @@ class ApplicationLogic(dynamo: DB, kong: Kong) {
   def createUser(form: CreateUserFormData): Future[ConsumerCreationResult] = {
     Logger.info(s"ApplicationLogic: Creating user with name ${form.name}")
     def saveUserAndKeyOnDB(consumer: ConsumerCreationResult, formData: CreateUserFormData, rateLimits: RateLimits): Unit = {
-      val newBonoboUser = BonoboUser(consumer.id, formData)
+      val labelIds = formData.labelIds match {
+        case Some(ids) => Some(ids.split(",").toList.filter(_.length > 0))
+        case None => None
```
cb372: Anywhere you are doing `case None => None`, I would recommend doing a `map` instead:

```scala
val labelIds = formData.labelIds.map(ids => ids.split(",").toList.filter(_.nonEmpty))
```

app/models/models.scala
```diff
@@ -13,28 +13,29 @@ case class BonoboUser(
   name: String,
   companyName: String,
   companyUrl: String,
-  additionalInfo: AdditionalUserInfo)
+  additionalInfo: AdditionalUserInfo,
+  labelIds: Option[List[String]])
```
cb372: I would make the list of labels a `List[String]` rather than `Option[List[String]]`. There's no reason to distinguish between `None` (i.e. no list of labels) and `Some(Nil)` (i.e. an empty list of labels).

cb372:
Looks good so far! 

We will also need the ability to create/update/delete the labels, but this could be a lot of work. Maybe best to manually add a few labels to Dynamo for now, and tackle the editing functionality in a separate PR.

Also, don't forget to add the new Dynamo table to CloudFormation.

cloudformation.json
```diff
@@ -340,6 +340,39 @@
...
+          }
+        ],
+        "KeySchema": [
+          {
+            "AttributeName":"hashkey",
+            "KeyType": "HASH"
+          }
+        ],
+        "ProvisionedThroughput":{
+          "ReadCapacityUnits": 20,
```
cb372: These capacity numbers are quite high and will cost :moneybag: 
Let's use 1 for both read and write, as Bonobo will only read the table once when it starts up.

app/store/Dynamo.scala
```diff
@@ -94,7 +100,11 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String) extends DB {
       new AttributeUpdate("email").put(bonoboUser.email),
       new AttributeUpdate("name").put(bonoboUser.name),
       new AttributeUpdate("companyName").put(bonoboUser.companyName),
-      new AttributeUpdate("companyName").put(bonoboUser.companyUrl)
+      new AttributeUpdate("companyName").put(bonoboUser.companyUrl),
+      bonoboUser.labelIds match {
+        case Nil => new AttributeUpdate("labelIds").delete()
+        case ids: List[String] => new AttributeUpdate("labelIds").put(ids.mkString(","))
```
cb372: It might be more sensible to model the list of IDs as a list in the DB, using Dynamo's `List` type rather than a comma-separated string.

app/store/Dynamo.scala
```diff
@@ -289,7 +308,11 @@ object Dynamo {
         commercialModel = Option(item.getString("commercialModel")),
         content = Option(item.getString("content")),
         contentFormat = Option(item.getString("contentFormat")),
-        articlesPerDay = Option(item.getString("articlesPerDay")))
+        articlesPerDay = Option(item.getString("articlesPerDay"))),
+      labelIds = Option(item.getString("labelIds")) match {
+        case Some(ids) => ids.split(",").toList
+        case None => List.empty
+      }
```
cb372: Indentation has gone a bit strange here. I'm surprised scalariform didn't fix that.

public/javascripts/createUser.js
```diff
@@ -0,0 +1,38 @@
...
+
+for(var i = 0; i < listLabels.length; i++)
+{
+    listLabels.item(i).addEventListener("click", function(){
+        var id = this.dataset.id
+        if(this.dataset.used == 'false') {
+            divChosenLabels.appendChild(this)
+            this.dataset.used = 'true'
+            hiddenLabelIds.value = hiddenLabelIds.value + "," + id
+            console.log(hiddenLabelIds.value)
```
cb372: Should probably delete the `console.log`s

public/javascripts/createUser.js
```diff
@@ -0,0 +1,38 @@
...
+        }
+        else if(this.dataset.used == 'true') {
+            divAllLabelsContainer.appendChild(this)
+            this.dataset.used = 'false'
+            hiddenLabelIds.value = hiddenLabelIds.value.replace(id, "")
+            console.log(hiddenLabelIds.value)
+        }
+    })
```
cb372: This JS looks reasonable to my amateur eyes, but it's worth testing in a few different browsers.

cb372:
Looks good apart from a few small comments.

Unfortunately we can't update the stack using CloudFormation because you added the missing index to the keys table and CloudFormation sucks :( Please create the new Dynamo table manually when you are ready.

PR #91, opened 2016-01-04
Fix hanging promise when using dummy email sending logic

cb372:
When we decide not to send an email (because we are in a pre-prod environment), we were not completing the promise, so the future never completed.


LATaylor-guardian:
:+1:

PR #92, opened 2016-01-05
Filter by labels

marialivia16:
- Filter keys page by user's labels
- Change pagination to use AJAX

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -33,19 +33,30 @@ class Application(dynamo: DB, kong: Kong, awsEmail: MailClient, labelsMap: Map[S
   def showKeys(direction: String, range: Option[String]) = maybeAuth { implicit request =>
     val resultsPage = dynamo.getKeys(direction, range)
     val totalKeys = dynamo.getNumberOfKeys()
-    val givenDirection = if (range.isDefined) direction else ""
-    Ok(views.html.showKeys(resultsPage.items, givenDirection, resultsPage.hasNext, totalKeys, request.user.firstName, pageTitle = "All Keys"))
+    Ok(views.html.showKeys(resultsPage.items, lastDirection = "", resultsPage.hasNext, totalKeys, labelsMap, request.user.firstName, pageTitle = "All Keys"))
+  }
+
+  def filter(withLabels: List[String], direction: String, range: Option[String]) = maybeAuth { implicit request =>
+    if (withLabels.isEmpty) {
+      val keys = dynamo.getKeys(direction, range)
+      val givenDirection = if (range.isDefined) direction else ""
+      Ok(views.html.renderKeysTable(keys.items, givenDirection, keys.hasNext))
+    } else {
+      val keys = dynamo.getKeys(direction, range, filterLabels = Some(withLabels))
+      val givenDirection = if (range.isDefined) direction else ""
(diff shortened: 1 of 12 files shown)


mchv:
@marialivia16 what is the benefit of using AJAX in this case? with the associated changes it seems to me that you break user history (i.e previous does not work correctly anymore) and error is not handled in the asynchronous call is not handled.

marialivia16:
@mchv Not using AJAX when paginating is causing the page to reload and lose the filtering conditions.

mchv:
@MahanaC you need to pass the filter in the query. Let discuss about this.

app/controllers/Application.scala
```diff
@@ -30,22 +30,28 @@ class Application(dynamo: DB, kong: Kong, awsEmail: MailClient, labelsMap: Map[S
 
   private val logic = new ApplicationLogic(dynamo, kong)
 
-  def showKeys(direction: String, range: Option[String]) = maybeAuth { implicit request =>
-    val resultsPage = dynamo.getKeys(direction, range)
+  def showKeys(labels: List[String], direction: String, range: Option[String]) = maybeAuth { implicit request =>
+    val keys = if (labels.isEmpty) dynamo.getKeys(direction, range) else dynamo.getKeys(direction, range, filterLabels = Some(labels))
```
cb372: This is nice and clear, but just for reference, a shorter way to write it would be

```scala
val keys = dynamo.getKeys(direction, range, Some(labels).filter(_.nonEmpty))
```

app/store/Dynamo.scala
```diff
@@ -27,11 +27,11 @@ trait DB {
    */
   def getUserWithEmail(email: String): Option[BonoboUser]
 
-  def saveKey(kongKey: KongKey): Unit
+  def saveKey(kongKey: KongKey, labelIds: List[String]): Unit
 
   def updateKey(kongKey: KongKey): Unit
```
cb372: Why doesn't this need to take a `labelIds` parameter? Did you implement the logic to update the labels on all of a user's keys when the user is updated?

cb372: Ah yes, I see you did implement it just below. Never mind.

app/store/Dynamo.scala
```diff
@@ -100,12 +102,21 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
...
-      new AttributeUpdate("companyName").put(bonoboUser.companyUrl),
+      new AttributeUpdate("companyUrl").put(bonoboUser.companyUrl),
       bonoboUser.labelIds match {
         case Nil => new AttributeUpdate("labelIds").delete()
         case ids: List[String] => new AttributeUpdate("labelIds").put(ids.asJava)
       }
     )
+    val keys = getKeysWithUserId(bonoboUser.bonoboId)
+    keys.foreach(kongKey =>
+      KongTable.updateItem(new PrimaryKey("hashkey", "hashkey", "rangekey", kongKey.rangeKey),
```
cb372: I would either extract this to a new method `private def updateKeyLabelIds(key: KongKey, labelIds: List[String])`, or add it to the `updateKey` method (and add a `labelIds` parameter to that method).

cb372:
:+1: apart from a couple of comments

PR #94, opened 2016-01-19
Add loader image to showKeys page

marialivia16:
Improves user experience when the filtering is being used.


mchv:
:+1: 

As a side note, rather than using a specific gif asset, the modern way will have [been to use css3](http://stackoverflow.com/a/26283602).

LATaylor-guardian:
@mchv Is such a software butterfly! ha

PR #95, opened 2016-01-24
[Security] Check that the user is guardian staff

mchv:
Not sure why [google-auth](https://github.com/guardian/play-googleauth) is not checking the domain, as I would have expecting it do the check, but [flexible code check the actual email address domain of the user](https://github.com/guardian/flexible-content/blob/d04957c030a00fb513046ff5f50a8b5e5f413c96/flexible-content-composer-backend/src/main/scala/com/gu/flexiblecontent/composerbackend/dispatcher/PanDomainAuthenticatedUser.scala#L215-L225) so do the same.

@cb372 @adamnfish @sihil


LATaylor-guardian:
:+1:

PR #96, opened 2016-01-25
Bump play-googleauth

cb372:
And remove temporary workaround

Google group and 2FA check to come in a later PR


LATaylor-guardian:
:+1:

PR #97, opened 2016-01-25
Add Google groups-based authorisation

cb372:
At the moment we only check that the user is in the two-factor auth group. 
Later we can add a requirement for the user to be in a whitelisted group of admin users.

I've created a Google service account to use for checking the user's group membership. The credentials for this service account are stored in a JSON file on S3 and downloaded when the machine starts up. ~~I've also asked for the email of the user we need to impersonate. Once I get a reply I will test this PR on CODE.~~ This has now been tested on CODE.


LATaylor-guardian:
:+1:

PR #99, opened 2016-02-08
Enable HTTPS on the ELB

cb372:
Also separate the CloudFormation templates for CODE and PROD


cb372:
This has been tested on CODE

LATaylor-guardian:
I think this is the best solution for now to get out of the mess we are in. :+1:

cb372:
Updated the PROD stack as well

PR #101, opened 2016-02-16
Improve Bonobo access logging

tomrf1:
1. Enable ELB access logs with CloudFormation
2. Add access logs to the Bonobo app, to provide better details about the user in a separate `audit.log` file.

The format of the new Bonobo log file is e.g.:
```
2016-02-16 17:39:20,056 - tom.forbes@guardian.co.uk GET /
2016-02-16 17:39:25,780 - tom.forbes@guardian.co.uk GET /key/search?query=test
2016-02-16 17:39:28,819 - tom.forbes@guardian.co.uk GET /user/<id>/edit
2016-02-16 17:39:31,852 - tom.forbes@guardian.co.uk POST /user/<id>/edit - Some(Map(csrfToken -> ArrayBuffer(<token>), name -> ArrayBuffer(Chris Test), email -> ArrayBuffer(chris.birchall+testtest@theguardian.com), companyName -> ArrayBuffer(The Guardian - testing), companyUrl -> ArrayBuffer(http://theguardian.com), labelIds -> ArrayBuffer()))
```

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -4,15 +4,15 @@ import email.MailClient
 import logic.ApplicationLogic
 import models._
 import com.gu.googleauth.{ UserIdentity, GoogleAuthConfig }
+import org.slf4j.LoggerFactory
 import play.api.data.Forms._
 import play.api.data._
 import play.api.i18n.{ I18nSupport, MessagesApi }
-import play.api.mvc.Security.AuthenticatedBuilder
+import play.api.mvc.Security.{ AuthenticatedRequest, AuthenticatedBuilder }
 import play.api.mvc._
 import store._
 import kong._
 import kong.Kong._
-import play.filters.csrf.CSRF
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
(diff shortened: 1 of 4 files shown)


mchv:
As a purely matter of style, you can do this with function composition putting the `audit` method inside the `maybeAuth` function but this is less flexible if you want to limit the audit to some functions not all of them using `maybeAuth`.

cb372:
I agree that we should put this inside the `maybeAuth` method if possible, so that we don't forget to add auditing to any new authenticated endpoints we add in the future.

This requires an understanding of Play's action composition architecture, which can be pretty mindbending... The [docs](https://www.playframework.com/documentation/2.4.x/ScalaActionsComposition) do not really help much. Mariot or I can help with this, although it always confuses me.

tomrf1:
I've updated this to use action composition, so that all request handlers that call `maybeAuth` will always perform auditing. This was...educational.

I still need to look into logging of authorisation failures, so i've added the not-ready-to-merge label.

LATaylor-guardian:
@tomrf1 If you can split the task into two: logging success and logging failure, then do that and it frees this up to be merged then. :)

PR #102, opened 2016-02-18
Update AMI

cb372:
to latest ubuntu-wily-java8

### DIFF
--- a/bonobo.service
+++ b/bonobo.service
@@ -6,7 +6,7 @@ User=content-api
 Group=content-api
 Restart=on-failure
 Environment='HOME=/home/content-api'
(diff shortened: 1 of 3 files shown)


LATaylor-guardian:
You could lower the instance size for this one while you're at it. (It's the only one left on that card after the explorer).

LATaylor-guardian:
:+1:

cb372:
OK, I'll make it t2.nano

cb372:
Note to self: I've updated the stack in CODE, still need to test.

cb372:
Seems stable in CODE after switching from a nano to a micro. I'll update the PROD stack and merge tomorrow.

cb372:
Updated PROD stack.

PR #105, opened 2016-02-29
Limit the size of all form inputs

marialivia16:


cb372:
:+1: thanks for doing this quite boring work!

PR #107, opened 2016-03-03
Set security and cache headers

marialivia16:
* Disable caching:

```
Expires: Tue, 03 Jul 2001 06:00:00 GMT
Last-Modified: {now} GMT
Cache-Control: no-store, no-cache, must-revalidate, max-age=0
Cache-Control: post-check=0, pre-check=0
Pragma: no-cache
```

* HSTS:

```
Strict-Transport-Security: max-age=31536000
```

* CSP (Content security policy)

* XSS:

```
X-XSS-Protection: 1; mode=block
```


mchv:
:+1: for reference disable caching is to protect personal data. It would be good as an improvement to disable this no-cache policy for assets.

PR #108, opened 2016-03-04
Add external tier to the model and views

marialivia16:


cb372:
:+1: (after accidentally merging it)

PR #109, opened 2016-03-08
Bump Play to 2.4.6; add CSRF vulnerability workaround

cb372:
* Bump Play to 2.4.6. This version includes playframework/playframework#5368 so we can remove a lot of the Play source that we copy-pasted.

* Add the workaround suggested in the [Play docs](https://www.playframework.com/security/vulnerability/20160304-CsrfBypass)


mchv:
:+1:

marialivia16:
It's nice that we can remove those copied files :smile:

cb372:
Tested on CODE.

PR #110, opened 2016-03-11
Use https url in the email

mchv:
This is dirty workaround but as `request.secure` is broken.


cb372:
#111 should make this unnecessary

mchv:
@cb372 great fix!

cb372:
It appears #111 did not work. Let's use this workaround for now.

PR #113, opened 2016-04-15
Set session.secure=false when running locally

cb372:
If we set it to true then the session becomes a secure cookie, which doesn't work because Play runs on http://localhost:9000

I've updated the CODE and PROD config files to overwrite it to true.


tsop14:
:+1:

PR #115, opened 2016-04-19
Make developer form fields optional

tsop14:
Makes company, companyUrl and productUrl fields optional on developer form - still required on commercial form. 
Adds test to check can create a user and key when those fields are left blank. 
Changes labels on UI - adds "(optional)" to optional fields and adds "or reason for key" to product field. 

screen shot 2016-04-19 at 15 21 50

cc @rrees

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -230,10 +230,10 @@ object Application {
     mapping(
       "name" -> nonEmptyText(maxLength = 200),
       "email" -> email.verifying("Maximum length is 200", _.length <= 200),
-      "companyName" -> nonEmptyText(maxLength = 200),
-      "companyUrl" -> nonEmptyText(maxLength = 200),
+      "companyName" -> optional(text(maxLength = 200)),
+      "companyUrl" -> optional(text(maxLength = 200)),
       "productName" -> nonEmptyText(maxLength = 200),
-      "productUrl" -> nonEmptyText(maxLength = 200),
+      "productUrl" -> optional(text(maxLength = 200)),
       "tier" -> nonEmptyText(maxLength = 200)
         .verifying(invalidTierMessage, tier => Tier.isValid(tier))
         .transform(tier => Tier.withNameOption(tier).get, (tier: Tier) => tier.toString),
@@ -248,8 +248,8 @@ object Application {
     mapping(
       "name" -> nonEmptyText(maxLength = 200),
       "email" -> email.verifying("Maximum length is 200", _.length <= 200),
-      "companyName" -> nonEmptyText(maxLength = 200),
-      "companyUrl" -> nonEmptyText(maxLength = 200),
+      "companyName" -> optional(text(maxLength = 200)),
+      "companyUrl" -> optional(text(maxLength = 200)),
       "labelIds" -> text(maxLength = 200)
     )(EditUserFormData.apply)(EditUserFormData.unapply)
   )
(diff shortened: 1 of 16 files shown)


marialivia16:
Is this also allowing people to leave the fields empty in the `user/create` endpoint?

tsop14:
It [is optional ](https://github.com/guardian/bonobo/pull/115/files#diff-84df020ba238a2b023209c399762d566R7) on the `user/create` endpoint.

app/controllers/Application.scala
```diff
@@ -230,10 +230,10 @@ object Application {
     mapping(
       "name" -> nonEmptyText(maxLength = 200),
       "email" -> email.verifying("Maximum length is 200", _.length <= 200),
-      "companyName" -> nonEmptyText(maxLength = 200),
-      "companyUrl" -> nonEmptyText(maxLength = 200),
+      "companyName" -> optional(text(maxLength = 200)),
```
marialivia16: Since all of these fields are optional on the endpoints I mentioned, it is worth adding the `(optional)` label in all the corresponding views.

marialivia16:
@tsop14 You can also update the description then.

app/controllers/DeveloperForm.scala
```diff
@@ -55,9 +55,9 @@ object DeveloperForm {
       "name" -> nonEmptyText(maxLength = 200),
       "email" -> email.verifying("Maximum length is 200", _.length <= 200),
       "productName" -> nonEmptyText(maxLength = 200),
-      "productUrl" -> nonEmptyText(maxLength = 200),
-      "companyName" -> nonEmptyText(maxLength = 200),
-      "companyUrl" -> nonEmptyText(maxLength = 200),
+      "productUrl" -> optional(text.verifying("Maximum length is 200", _.length <= 200)),
```
marialivia16: You can use `optional(text(maxLength = 200))` like you do in `Application.scala`.

app/email/AwsEmailClient.scala
```diff
@@ -10,7 +10,7 @@ import play.api.mvc.RequestHeader
 import scala.concurrent.{ Future, Promise }
 
 trait MailClient {
-  def sendEmailCommercialRequestToModerators(user: BonoboUser, productName: String, productUrl: String)(implicit request: RequestHeader): Future[SendEmailResult]
+  def sendEmailCommercialRequestToModerators(user: BonoboUser, productName: String, productUrl: Option[String])(implicit request: RequestHeader): Future[SendEmailResult]
```
cb372: Shouldn't the `productUrl` still be `String` here, as all fields are mandatory for commercial requests?

app/store/Dynamo.scala
```diff
@@ -308,12 +317,12 @@ object Dynamo {
...
       .withString("name", bonoboKey.name)
       .withString("email", bonoboKey.email)
-      .withString("companyName", bonoboKey.companyName)
-      .withString("companyUrl", bonoboKey.companyUrl)
       .withLong("createdAt", bonoboKey.additionalInfo.createdAt.getMillis)
       .withString("registrationType", bonoboKey.additionalInfo.registrationType.friendlyName)
       .withList("labelIds", bonoboKey.labelIds.asJava)
 
+    bonoboKey.companyName.fold(item) { companyName => item.withString("companyName", companyName) }
+    bonoboKey.companyUrl.fold(item) { companyUrl => item.withString("companyUrl", companyUrl) }
```
cb372: Unrelated to your change, which is fine, but this pattern of throwing away the result of the `fold` and relying on the fact that `Item` is mutable makes me a bit queasy... 😷

cb372:
Looks good to me 👍

PR #118, opened 2016-06-29
Update to Play 2.4.8

tomrf1:
deployed to CODE


cb372:
👍 

By the way, when we upgrade Bonobo to 2.5.x we should be careful we don't break anything related to https://www.playframework.com/security/vulnerability/20160304-CsrfBypass

PR #119, opened 2016-07-27
Add endpoint to get all user emails

oilnam:
Creates endpoint to get all registered emails: `/email/$tier/$status`

E.g.

https://bonobo-code.capi.gutools.co.uk/email/developer/active
https://bonobo-code.capi.gutools.co.uk/email/internal/inactive

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -8,6 +8,7 @@ import org.slf4j.LoggerFactory
 import play.api.data.Forms._
 import play.api.data._
 import play.api.i18n.{ I18nSupport, MessagesApi }
+import play.api.libs.json.Json
 import play.api.mvc.Security.{ AuthenticatedRequest, AuthenticatedBuilder }
 import play.api.mvc._
 import store._
@@ -209,6 +210,10 @@ class Application(dynamo: DB, kong: Kong, awsEmail: MailClient, labelsMap: Map[S
     }
 
     editKeyForm.bindFromRequest.fold[Future[Result]](handleInvalidForm, handleValidForm)
+  }
+
+  def getEmails(tier: String, status: String) = maybeAuth { implicit request =>
+    Ok(Json.toJson(dynamo.getEmails(tier, status)))
   }
 
   def healthcheck = Action { Ok("OK") }
--- a/app/store/Dynamo.scala
+++ b/app/store/Dynamo.scala
@@ -51,6 +51,8 @@ trait DB {
   def getLabels(): List[Label]
 
   def getLabelsFor(bonoboId: String): List[String]
+
+  def getEmails(tier: String, status: String): List[String]
 }
 
 class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: String) extends DB {
@@ -296,6 +298,27 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
     }
   }
 
+  def getEmails(tier: String, status: String): List[String] = {
+    val query = new ScanSpec()
+      .withFilterExpression("#sts = :s AND tier = :t")
+      .withNameMap(new NameMap()
+        .`with`("#sts", "status")
+      )
+      .withValueMap(new ValueMap().withString(":s", status.capitalize).withString(":t", tier.capitalize))
+
+    val userIds = KongTable.scan(query).asScala.toList.map(_.getString("bonoboId"))
+
+    def getEmailForUser(id: String): Option[String] = {
+      val userQuery = new QuerySpec()
+        .withKeyConditionExpression("id = :i")
+        .withValueMap(new ValueMap().withString(":i", id))
+        .withMaxResultSize(1)
+      BonoboTable.query(userQuery).asScala.toList.map(_.getString("email")).headOption
+    }
+
+    userIds.flatMap(getEmailForUser)
+  }
+
   /**
    * The following methods are used for labeling an user
    */
@@ -308,6 +331,7 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
       .withValueMap(new ValueMap().withString(":i", bonoboId))
       .withMaxResultSize(1)).asScala.toList.map(fromBonoboItem).flatMap(_.labelIds)
   }
+
 }
 
 object Dynamo {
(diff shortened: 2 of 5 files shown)


cb372:
Didn't we want to filter out users that don't have any active keys?

app/controllers/Application.scala
```diff
@@ -211,6 +212,10 @@ class Application(dynamo: DB, kong: Kong, awsEmail: MailClient, labelsMap: Map[S
     editKeyForm.bindFromRequest.fold[Future[Result]](handleInvalidForm, handleValidForm)
   }
 
+  def getAllEmails = maybeAuth { implicit request =>
```
LATaylor-guardian: This function name could change now given the context of your change.

app/store/Dynamo.scala
```diff
@@ -296,6 +298,23 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
     }
   }
 
+  def getAllActiveDeveloperEmails(): List[String] = {
+    val query = new ScanSpec()
+      .withFilterExpression("status = :s").withValueMap(new ValueMap().withString(":s", "Active"))
+      .withFilterExpression("tier = :t").withValueMap(new ValueMap().withString(":t", "Developer"))
+    val userIds = KongTable.scan(query).asScala.toList.map(_.getString("bonoboId"))
+
+    def getEmailForUser(id: String): String = {
```
LATaylor-guardian: Is there any way to do this without making a query per userId?

conf/routes
```diff
@@ -18,6 +18,8 @@ POST        /user/create                controllers.Application.createUser
 GET         /user/:id/edit              controllers.Application.editUserPage(id: String)
 POST        /user/:id/edit              controllers.Application.editUser(id: String)
 
+GET         /email/all                  controllers.Application.getAllEmails
```
LATaylor-guardian: If you want this to be more restful, you can simply have `/email` to return all emails. Then to filter you would use query params such as `activeDeveloper=true`. Again, this is a trade off between creating something more beautiful or putting up an endpoint for Chris to use quickly :)

app/store/Dynamo.scala
```diff
@@ -296,6 +298,23 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
...
+      val userQuery = new QuerySpec()
+        .withKeyConditionExpression("id = :i")
+        .withValueMap(new ValueMap().withString(":i", id))
+        .withMaxResultSize(1)
+      BonoboTable.query(userQuery).asScala.toList.map(_.getString("email")).head
```
LATaylor-guardian: It might be better to use `headOption` here rather than `head`.

The line below `userIds.map(getEmailForUser)` can then become `userIds.flatMap(getEmailForUser)' and we then don't risk trying to access an elements that isn't there if the query changes in the future to not have a maxResultsSize of 1.

app/store/Dynamo.scala
```diff
@@ -321,15 +341,15 @@ object Dynamo {
       .withString("registrationType", bonoboKey.additionalInfo.registrationType.friendlyName)
       .withList("labelIds", bonoboKey.labelIds.asJava)
 
-    bonoboKey.companyName.foreach( companyName => item.withString("companyName", companyName) )
-    bonoboKey.companyUrl.foreach( companyUrl => item.withString("companyUrl", companyUrl) )
+    bonoboKey.companyName.foreach(companyName => item.withString("companyName", companyName))
+    bonoboKey.companyUrl.foreach(companyUrl => item.withString("companyUrl", companyUrl))
```
mchv: why removing the whitespace?

app/store/Dynamo.scala
```diff
@@ -321,15 +341,15 @@ object Dynamo {
...
-    bonoboKey.additionalInfo.commercialModel.foreach( commercialModel => item.withString("commercialModel", commercialModel) )
-    bonoboKey.additionalInfo.content.foreach( content => item.withString("content", content) )
-    bonoboKey.additionalInfo.contentFormat.foreach( contentFormat => item.withString("contentFormat", contentFormat) )
-    bonoboKey.additionalInfo.articlesPerDay.foreach( articlesPerDay => item.withString("articlesPerDay", articlesPerDay) )
+    bonoboKey.additionalInfo.businessArea.foreach(businessArea => item.withString("businessArea", businessArea))
+    bonoboKey.additionalInfo.monthlyUsers.foreach(monthlyUsers => item.withString("monthlyUsers", monthlyUsers))
+    bonoboKey.additionalInfo.commercialModel.foreach(commercialModel => item.withString("commercialModel", commercialModel))
+    bonoboKey.additionalInfo.content.foreach(content => item.withString("content", content))
+    bonoboKey.additionalInfo.contentFormat.foreach(contentFormat => item.withString("contentFormat", contentFormat))
+    bonoboKey.additionalInfo.articlesPerDay.foreach(articlesPerDay => item.withString("articlesPerDay", articlesPerDay))
```
mchv: idem

app/store/Dynamo.scala
```diff
@@ -372,7 +392,7 @@ object Dynamo {
       .withString("productName", kongKey.productName)
       .withList("labelIds", labelIds.asJava)
 
-    kongKey.productUrl.foreach( productUrl => item.withString("productUrl", productUrl) )
+    kongKey.productUrl.foreach(productUrl => item.withString("productUrl", productUrl))
```
mchv: idem

app/store/Dynamo.scala
oilnam: @mchv I didn't do it, `scalac` does it automatically when you run the tests.

LATaylor-guardian: image

app/store/Dynamo.scala
```diff
@@ -296,6 +298,20 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
...
+  def getEmails(tier: String, status: String): List[String] = {
+    val query = new ScanSpec()
+      .withFilterExpression("#sts = :s AND tier = :t")
+      .withNameMap(new NameMap()
+        .`with`("#sts", "status")
+      )
+      .withValueMap(new ValueMap().withString(":s", status.capitalize).withString(":t", tier.capitalize))
+
+    val userIds = KongTable.scan(query).asScala.toList.map(_.getString("bonoboId"))
+    val allUsers = BonoboTable.scan(new ScanSpec).asScala.toList.map(fromBonoboItem)
```
LATaylor-guardian: if when mapping over this you turn it into a map, you can access in constant time below rather than having to traverse the list. Upto you whether you care!

LATaylor-guardian:
👍

app/store/Dynamo.scala
oilnam: Not really, the list is small and the bottleneck would probably be making the dynamo request anyway. Good point though! ;)

Issue #112, opened 2016-04-10
docker-compose instead of configure-docker.sh


arahayrabedian:
I see a chunk of stuff in ```configure-docker.sh``` that could easily be done cleanly with docker-compose, I haven't used this service yet but was just reading about it thought I'd drop that recommendation.

cb372:
Thanks for the tip. I'll look into migrating to docker-compose next time we need to do some work on this project.

PR #121, opened 2016-08-25
docker-compose-ify dev bootstrap scripts

arahayrabedian:
addresses #112 by internally changing the `scripts/configure-docker.sh` script to use `docker-compose` and `docker-compose.yml`. It's not the cleanest solution (I think when the new `HEALTHCHECK` directive has better support this can be made much cleaner) but it works locally - which I believe is the purpose here.

I don't know how to test it further than those two POST requests returning 201's.

I often see a warning appear, the consequences of which I'm unfamiliar with:

> kong_1       | [WARN] Cannot auto-join the cluster because no nodes were found

although the whole thing still seems to work.

You may also be interested in more advanced setups by the kong guys themselves [here](https://github.com/Mashape/docker-kong/blob/master/compose/docker-compose.yml)

If there needs to be some sort of persistence, I can try to bake that in too but I'm not too familiar with cassandra.

### DIFF
--- a/scripts/configure-docker.sh
+++ b/scripts/configure-docker.sh
@@ -1,10 +1,14 @@
 #! /bin/bash
 
+docker_host=${DOCKER_HOST?Docker host is not set}
+
+container_host=$(echo ${docker_host} | sed -e 's/tcp:\/\/\(.*\):.*/\1/')
+
 echo Starting Services ...
 docker-compose up -d
 
 echo Adding API ...
-wget -O - http://127.0.0.1:8001/apis --post-data 'name=internal&request_host=foo.com&upstream_url=http://example.com' --retry-connrefused --no-verbose
+wget -O - http://${container_host}:8001/apis --post-data 'name=internal&request_host=foo.com&upstream_url=http://example.com' --retry-connrefused --no-verbose
 
 echo Activating key-auth plugin ...
-wget -O - http://127.0.0.1:8001/apis/internal/plugins/ --post-data 'name=key-auth' --retry-connrefused --no-verbose
+wget -O - http://${container_host}:8001/apis/internal/plugins/ --post-data 'name=key-auth' --retry-connrefused --no-verbose
--- a/scripts/configure-docker.sh
+++ b/scripts/configure-docker.sh
@@ -1,45 +1,10 @@
 #! /bin/bash
 
-docker_host=${DOCKER_HOST?Docker host is not set}
-
-container_host=$(echo ${docker_host} | sed -e 's/tcp:\/\/\(.*\):.*/\1/')
-
-function start_service {
-    name=${1?Name parameter missing}
-    port_number=${2?Port parameter missing}
-
-    docker start ${name}
-
-    nc -z ${container_host} ${port_number}
-    while [ $? -ne 0 ]; do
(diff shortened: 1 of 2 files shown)


arahayrabedian:
oop, my assumption of `127.0.0.1` snuck in there, will fix that shortly, need to run from the computer for a bit.

cb372:
Awesome! We were just saying yesterday that we should get around to doing this, but you beat us to it.

I wouldn't worry about Cassandra persistence.

scripts/configure-docker.sh
```diff
@@ -1,45 +1,10 @@
...
-
-start_service cassandra 9042
-
-start_service kong 8001
+echo Starting Services ...
+docker-compose up -d
 
 echo Adding API ...
-curl -sS -X POST http://${container_host}:8001/apis -d name=internal -d request_host=foo.com -d upstream_url=http://example.com
+wget -O - http://127.0.0.1:8001/apis --post-data 'name=internal&request_host=foo.com&upstream_url=http://example.com' --retry-connrefused --no-verbose
```
cb372: Any reason for the switch from curl to wget?

arahayrabedian: yeah, `curl` does not consider a refused connection as transient and thus never retries, wget has the `--retry-connrefused` option. Personally was disappointed that curl couldn't do it (or a quick search did not reveal a way)

the curl man page says of the `--retry` option:
>Transient error means either: a timeout, an FTP 4xx code or an HTTP 5xx response code.

so a connection refusal still exits without retrying.

arahayrabedian:
@cb372 I'm having issues when i set `DOCKER_HOST` on my machine, I'm guessing you guys are using docker-machine?

I restored it to how it was, but I'm having trouble giving it a spin (as I did when I ran the original script)

arahayrabedian:
hey guys, let me know when you're done with the stuff in #122 and i'll get this up to speed. Alternatively, let me know what the final state should be and I can work off that info.

LATaylor-guardian:
Hey @arahayrabedian, we are planning on migration from a Cassandra backed Kong, to a Postgres backed Kong, so things will be in a state of flux for the time being, so I will keep you updated.

LATaylor-guardian:
Hey @arahayrabediant things should be in a good state for you to pick this back up again. Many thanks.

arahayrabedian:
cool @LATaylor-guardian , i'll cobble something together in the next few days, still settling down (just moved) and binging on a series all at once.

arahayrabedian:
that took longer than expected to actually get running... turns out google DNS servers are blocked on campus - is that a UK-wide thing?

anyway, methinks I'm done here - not a fan of how I have to keep retrying at two different points, but it really does work.

arahayrabedian:
hey @LATaylor-guardian, can you guys give this a spin and let me know if you face any issues?

LATaylor-guardian:
@arahayrabedian Will do this today! Many thanks.

LATaylor-guardian:
@arahayrabedian Sorry, didn't get round to this - will pick up first thing tomorrow.

arahayrabedian:
hey @LATaylor-guardian there's no hurry, just hoping to get it off my plate before the semester picks up.

PR #122, opened 2016-08-25
Write to both the old and new Kong instance simultaneously.

LATaylor-guardian:
Tests will fail as I have yet to tackle them and docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+
 # bonobo
 
 [![Circle CI](https://circleci.com/gh/guardian/bonobo/tree/master.svg?style=svg)](https://circleci.com/gh/guardian/bonobo/tree/master)
@@ -12,6 +13,8 @@ Bonobo needs an instance of Kong to connect to, so you will need Kong and Cassan
 
 Install Docker Toolkit, following the instructions on the [Docker website](http://docs.docker.com/).
 
+Make sure you have [virtualbox](https://www.virtualbox.org/) installed.
+
 1. Make sure your Docker VM is running:
 
   ```
@@ -19,6 +22,13 @@ Install Docker Toolkit, following the instructions on the [Docker website](http:
   NAME   ACTIVE   DRIVER       STATE     URL                         SWARM
   dev    *        virtualbox   Running   tcp://192.168.99.100:2376
   ```
+
+1a. If nothing is listed here, you will need to create it.
+
+  ```
+  $ docker-machine create --driver virtualbox dev
+  ```
+
 
 2. Make sure `$DOCKER_HOST` is set:
 
--- a/README.md
+++ b/README.md
@@ -22,13 +22,12 @@ Make sure you have [virtualbox](https://www.virtualbox.org/) installed.
   NAME   ACTIVE   DRIVER       STATE     URL                         SWARM
   dev    *        virtualbox   Running   tcp://192.168.99.100:2376
   ```
-
-1a. If nothing is listed here, you will need to create it.
-
+  
+  If nothing is listed here, you will need to create it.
+  
   ```
   $ docker-machine create --driver virtualbox dev
   ```
-
 
 2. Make sure `$DOCKER_HOST` is set:
 
--- a/app/kong/Kong.scala
+++ b/app/kong/Kong.scala
@@ -51,13 +51,13 @@ object Kong {
 trait Kong {
   import Kong._
 
-  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: Option[String]): Future[ConsumerCreationResult]
+  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: String): Future[ConsumerCreationResult] // TODO - make key an option again after migration
 
   def updateConsumer(id: String, newRateLimit: RateLimits): Future[Happy.type]
 
   def updateConsumerUsername(id: String, tier: Tier): Future[Happy.type]
 
-  def createKey(consumerId: String, customKey: Option[String] = None): Future[String]
+  def createKey(consumerId: String, customKey: String): Future[String] // TODO - make key an option again after migration
 
   def deleteKey(consumerId: String): Future[Happy.type]
 }
@@ -69,7 +69,7 @@ class KongClient(ws: WSClient, serverUrl: String, apiName: String) extends Kong 
   val RateLimitingPluginName = "rate-limiting"
   val KeyAuthPluginName = "key-auth"
 
-  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: Option[String]): Future[ConsumerCreationResult] = {
+  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: String): Future[ConsumerCreationResult] = {
     for {
       consumer <- createConsumer(tier)
       _ <- setRateLimit(consumer.id, rateLimit)
@@ -109,8 +109,8 @@ class KongClient(ws: WSClient, serverUrl: String, apiName: String) extends Kong 
     }
   }
 
-  def createKey(consumerId: String, customKey: Option[String] = None): Future[String] = {
-    val key: String = customKey getOrElse java.util.UUID.randomUUID.toString
+  def createKey(consumerId: String, customKey: String): Future[String] = {
+    val key: String = customKey // TODO make custom key an option again after migration
     ws.url(s"$serverUrl/consumers/$consumerId/$KeyAuthPluginName").post(Map(
       "key" -> Seq(key))).flatMap {
       response =>
--- a/app/kong/Kong.scala
+++ b/app/kong/Kong.scala
@@ -51,13 +51,13 @@ object Kong {
 trait Kong {
   import Kong._
 
-  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: String): Future[ConsumerCreationResult] // TODO - make key an option again after migration
+  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: String): Future[ConsumerCreationResult]
 
   def updateConsumer(id: String, newRateLimit: RateLimits): Future[Happy.type]
 
   def updateConsumerUsername(id: String, tier: Tier): Future[Happy.type]
 
-  def createKey(consumerId: String, customKey: String): Future[String] // TODO - make key an option again after migration
+  def createKey(consumerId: String, customKey: String): Future[String]
 
   def deleteKey(consumerId: String): Future[Happy.type]
 }
@@ -110,7 +110,7 @@ class KongClient(ws: WSClient, serverUrl: String, apiName: String) extends Kong 
   }
 
   def createKey(consumerId: String, customKey: String): Future[String] = {
-    val key: String = customKey // TODO make custom key an option again after migration
+    val key: String = customKey
     ws.url(s"$serverUrl/consumers/$consumerId/$KeyAuthPluginName").post(Map(
       "key" -> Seq(key))).flatMap {
       response =>
--- a/app/kong/KongWrapper.scala
+++ b/app/kong/KongWrapper.scala
@@ -19,33 +19,51 @@ case class KongWrapper(existingKong: Kong, newKong: Kong) {
     }
   }
 
-  def createKey(consumerId: String, customKey: Option[String] = None): Future[KongKeyWrapper] = {
-    for {
-      key <- existingKong.createKey(consumerId, customKey)
-      migrationKey <- newKong.createKey(consumerId, customKey)
-    } yield {
-      KongKeyWrapper(key, migrationKey)
+  def createKey(consumerId: String, maybeMigrationKongId: Option[String], customKey: Option[String] = None): Future[KongKeyWrapper] = {
+    maybeMigrationKongId match {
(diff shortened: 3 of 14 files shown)


app/kong/KongWrapper.scala
```diff
@@ -1,3 +1,51 @@
 package kong
 
-case class KongWrapper(existingKong: Kong, newKong: Kong)
+import kong.Kong.Happy
```
cb372: I'd forgotten all about good old `Happy`

cb372:
Will [this line](https://github.com/guardian/bonobo/blob/e80ca5012bf4cf26936621204019d8a33578d38e/app/kong/Kong.scala#L114) work in Kong 0.9? Didn't you say this endpoint was changed?

LATaylor-guardian:
No idea yet, will need to test/check.

app/kong/KongWrapper.scala
```diff
@@ -1,3 +1,51 @@
...
+
+case class KongKeyWrapper(key: String, migrationKey: String)
+case class ConsumerCreationResultWrapper(consumerCR: ConsumerCreationResult, migrationConsumerCR: ConsumerCreationResult)
+
+case class KongWrapper(existingKong: Kong, newKong: Kong) {
+
+  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: Option[String]): Future[ConsumerCreationResultWrapper] = {
+    for {
+      cr1 <- existingKong.createConsumerAndKey(tier, rateLimit, key)
+      cr2 <- newKong.createConsumerAndKey(tier, rateLimit, key)
```
cb372: These will happen in serial and will fail-fast if the first one fails. Is that what you intended? Same goes for all of the other methods below.

LATaylor-guardian: Yup, I see no reason to attempt the second call if the first one fails.

app/kong/KongWrapper.scala
```diff
@@ -1,3 +1,51 @@
...
+  def createKey(consumerId: String, customKey: Option[String] = None): Future[KongKeyWrapper] = {
+    for {
+      key <- existingKong.createKey(consumerId, customKey)
+      migrationKey <- newKong.createKey(consumerId, customKey)
+    } yield {
+      KongKeyWrapper(key, migrationKey)
+    }
+  }
+
+  def updateConsumerUsername(consumerId: String, tier: Tier): Future[Happy.type] = {
```
cb372: The consumerId will be different for each Kong (and may not exist on the new Kong)

app/kong/KongWrapper.scala
```diff
@@ -1,3 +1,51 @@
...
+    for {
+      _ <- existingKong.updateConsumerUsername(consumerId, tier)
+      _ <- newKong.updateConsumerUsername(consumerId, tier)
+    } yield Happy
+  }
+
+  def updateConsumer(consumerId: String, newRateLimit: RateLimits): Future[Happy.type] = {
```
cb372: Same as above

app/kong/KongWrapper.scala
```diff
@@ -1,3 +1,51 @@
...
+  def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: Option[String]): Future[ConsumerCreationResultWrapper] = {
+    for {
+      cr1 <- existingKong.createConsumerAndKey(tier, rateLimit, key)
+      cr2 <- newKong.createConsumerAndKey(tier, rateLimit, key)
+    } yield {
+      ConsumerCreationResultWrapper(cr1, cr2)
+    }
+  }
+
+  def createKey(consumerId: String, customKey: Option[String] = None): Future[KongKeyWrapper] = {
```
cb372: Same here, consumerId will differ between Kongs

app/kong/KongWrapper.scala
```diff
@@ -1,3 +1,51 @@
...
+  }
+
+  def updateConsumer(consumerId: String, newRateLimit: RateLimits): Future[Happy.type] = {
+    for {
+      _ <- existingKong.updateConsumer(consumerId, newRateLimit)
+      _ <- newKong.updateConsumer(consumerId, newRateLimit)
+    } yield Happy
+  }
+
+  def deleteKey(consumerId: String): Future[Happy.type] = {
```
cb372: and here

app/kong/KongWrapper.scala
LATaylor-guardian: Yeah, sorry, I forgot to add I haven't changed the `retrievingKeyFromDynamo` function yet to return both keys.

app/kong/Kong.scala
```diff
@@ -109,8 +109,8 @@ class KongClient(ws: WSClient, serverUrl: String, apiName: String) extends Kong
     }
   }
 
-  def createKey(consumerId: String, customKey: Option[String] = None): Future[String] = {
-    val key: String = customKey getOrElse java.util.UUID.randomUUID.toString
+  def createKey(consumerId: String, customKey: String): Future[String] = {
+    val key: String = customKey // TODO make custom key an option again after migration
```
cb372: I think we should probably leave it like this (pass in a `String` rather than `Option[String]`) after the migration. It seems strange that the `Kong` client is in charge of generating a key - this should be done further up the stack.

app/kong/KongWrapper.scala
```diff
@@ -1,3 +1,73 @@
 package kong
 
-case class KongWrapper(existingKong: Kong, newKong: Kong)
+import kong.Kong.Happy
+import models.{ ConsumerCreationResult, RateLimits, Tier }
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+case class KongKeyWrapper(key: String, migrationKey: String)
```
cb372: It doesn't really matter because the contents of this class are never read, but `migrationKey` should be an `Option[String]`.

cb372:
I would call the `migrationKongId` field `kongConsumerId` in Dynamo. So once migration and cleanup are finished, we'll be left with a usefully-named field, and the confusingly-named `kongId` field will be gone.

README.md
```diff
@@ -20,6 +23,13 @@ Install Docker Toolkit, following the instructions on the [Docker website](http:
   dev    *        virtualbox   Running   tcp://192.168.99.100:2376
   ```
 
+1a. If nothing is listed here, you will need to create it.
```
cb372: This `1a.` will mess up the markdown numbering

LATaylor-guardian: gah.

LATaylor-guardian:
> I would call the migrationKongId field kongConsumerId in Dynamo. So once migration and cleanup are finished, we'll be left with a usefully-named field, and the confusingly-named kongId field will be gone.

Agreed. The only annoying thing about this is that the kongConsumerId becomes an Option after a migration in our data model. Annoying.

app/logic/ApplicationLogic.scala
```diff
@@ -88,11 +88,17 @@ class ApplicationLogic(dynamo: DB, kong: KongWrapper) {
     Logger.info(s"ApplicationLogic: Creating key for user with id $userId")
     def createConsumerAndKey: Future[String] = {
       val rateLimits: RateLimits = form.tier.rateLimit
-      kong.existingKong.createConsumerAndKey(form.tier, rateLimits, form.key) flatMap {
+      kong.createConsumerAndKey(form.tier, rateLimits, form.key) flatMap {
         consumer =>
           {
-            saveKeyOnDB(userId, consumer, rateLimits, form.tier, form.productName, form.productUrl, dynamo.getLabelsFor(userId))
-            Future.successful(consumer.key)
+            /**
```
cb372: I don't understand this logic. The procedure for adding a key to an existing user should be as follows:

1. Create consumer and key in both Kongs
2. Save the key to Dynamo, with the userId, old Kong consumerId and new Kong consumerId

I think the line that looks up the consumerId from existing keys in Dynamo is incorrect.

app/logic/ApplicationLogic.scala
```diff
@@ -109,6 +115,7 @@ class ApplicationLogic(dynamo: DB, kong: KongWrapper) {
     Logger.info(s"ApplicationLogic: Updating key with id ${oldKey.kongId}")
     val bonoboId = oldKey.bonoboId
     val kongId = oldKey.kongId
+    val maybeMigrationKongId = oldKey.kongConsumerId
 
     def updateKongKeyOnDB(form: EditKeyFormData): Unit = {
       val updatedKey = {
```
cb372: The two calls to `KongKey(...)` in this block are passing `None` as the `migrationKongId` param, so the new consumer ID will be lost from Dynamo.

app/logic/ApplicationLogic.scala
```diff
@@ -122,31 +129,35 @@ class ApplicationLogic(dynamo: DB, kong: KongWrapper) {
...
     }
 
+    /**
+     * Do not use this function anywhere other than the invocation below. until after the migration.
+     */
     def activateKeyIfNecessary(): Future[String] = {
       if (oldKey.status == KongKey.Inactive && form.status == KongKey.Active) {
-        kong.existingKong.createKey(kongId, Some(oldKey.key))
+        kong.createKey(kongId, maybeMigrationKongId, Some(oldKey.key))
+        Future.successful("") // return value doesn't currently matter as it's not being used anywhere other than to be used within a for comprehension where it's thrown away.
```
cb372: This is a bit confusing, as you spawn one future to create the key but return another future. So this method's result will complete while the key creation future is still running.

As the method's result is not used anywhere, why not change its return type to `Future[Unit]`?

cb372:
Unrelated to your changes, but I just noticed that the integration tests don't do any checks on rate limits being set in Kong (except the test for altering rate limits). Would be nice to add this.

integration-test/scala/integration/IntegrationTests.scala
```diff
@@ -385,14 +417,22 @@ class IntegrationTests extends FlatSpec with Matchers with OptionValues with Int
...
     // check Kong's key value matches Bonobo-Keys.keyValue
     val keyValue = Await.result(getKeyForConsumerId(consumerId), atMost = 10.seconds)
+    val migrationKeyValue = Await.result(getKeyForConsumerIdOnMigrationKong(migrationConsumerId), atMost = 10.seconds)
+
     val dynamoKongKey = dynamo.getKeyWithValue(keyValue)
+    val migrationDynamoKongKey = dynamo.getKeyWithValue(migrationKeyValue)
+
     keyValue shouldBe dynamoKongKey.value.key
+    migrationKeyValue shouldBe migrationDynamoKongKey.value.key
+
```
cb372: Could add a check that `keyValue == migrationKeyValue`, i.e. the same key has been created on both Kongs. This check could also be added to a few other tests.

app/logic/ApplicationLogic.scala
LATaylor-guardian: Good spot, missed that function.

integration-test/scala/integration/IntegrationTests.scala
```diff
@@ -178,9 +198,10 @@ class IntegrationTests extends FlatSpec with Matchers with OptionValues with Int
       bonoboId = consumerId,
       additionalInfo = userToSave.additionalInfo.copy(createdAt = dynamoBonoboUser.value.additionalInfo.createdAt))
 
-    dynamoKongKey.value shouldBe keyToSave.copy(
+    dynamoKongKey.value.copy(kongConsumerId = migrationConsumerId) shouldBe keyToSave.copy(
```
cb372: I don't think you need the `.copy(kongConsumerId = migrationConsumerId)` here. If you do, it probably means something is wrong in the production code.

integration-test/scala/integration/IntegrationTests.scala
```diff
@@ -220,15 +242,18 @@ class IntegrationTests extends FlatSpec with Matchers with OptionValues with Int
...
     // check the consumerId in dynamo matches the one on Kong
     Await.result(checkConsumerExistsOnKong(secondKongId), atMost = 10.seconds) shouldBe true
+    Await.result(checkConsumerExistsOnMigrationKong(migrationSecondKongId), atMost = 10.seconds) shouldBe true
 
     bonoboId shouldBe firstKongId
     bonoboId should not be secondKongId
 
     // the bonoboId for the new key should be same as the bonoboId for the first one
     bonoboId shouldBe dynamo.getKeyWithValue("the-dark-day").value.bonoboId
+    migrationKongId shouldBe dynamo.getKeyWithValue("the-dark-day").value.kongConsumerId
```
cb372: This assertion is incorrect. No two keys should have the same consumerId.

integration-test/scala/integration/IntegrationTests.scala
```diff
@@ -269,6 +296,7 @@ class IntegrationTests extends FlatSpec with Matchers with OptionValues with Int
 
     // check the key doesn't exist on Kong anymore
     Await.result(checkKeyExistsOnKong(consumerId), atMost = 10.seconds) shouldBe false
+    Await.result(checkKeyExistsOnMigrationKong(migrationConsumerId), atMost = 10.seconds) shouldBe false
```
cb372: Could add a check that `kongConsumerId` is still set correctly on the Dynamo record at this point. This would catch the issue I commented on above, where it was accidentally being set to `None`.

PR #124, opened 2016-09-23
Bonobo updates after kong migration

LATaylor-guardian:


app/components/AppComponents.scala
```diff
@@ -81,27 +81,17 @@ trait DynamoComponentImpl extends DynamoComponent { self: BuiltInComponents =>
...
-    }
-
-    val newKong = {
-      val apiAddress = confString("kong.new.apiAddress")
-      val apiName = confString("kong.new.apiName")
-      new KongClient(wsClient, apiAddress, apiName)
-    }
-
-    KongWrapper(existingKong, newKong)
+    val apiAddress = confString("kong.existing.apiAddress")
```
LATaylor-guardian: Will change in a seperate PR after.

app/models/models.scala
```diff
@@ -57,8 +57,8 @@ object AdditionalUserInfo {
 /* model used for saving the keys on Kong */
 case class KongKey(
   bonoboId: String,
-  kongId: String,
-  migrationKongId: Option[String], // kong ID for the consumer created against the Kong stack we are migrating to.
+  kongId: Option[String], // to be deleted.
```
LATaylor-guardian: This has become an option to enable it to be deleted. This is done using scripts I have written to remove the attribute on CODE. I have tested this on CODE and removed the attribute and it works as expected.

mchv:
👍

PR #125, opened 2016-09-26
Bonobo - remove kongId from KongKey model.

LATaylor-guardian:
- Now that the kong migration is complete, the kongId attribute has been removed
from the keys table in Dynamo for all keys (the canonical source for the consumer id provided by Kong is now kongConsumerId). This means the kongId field can be removed from the KongKey model within Bonobo.


mchv:
👍  good work

PR #126, opened 2016-12-05
yaml cloudformation

tomrf1:
The 2 files were/are identical, except the CODE one doesn't have the dynamo table definitions.

Tested in CODE.


LATaylor-guardian:
@tomrf1 Yes, we have two seperate files for good reason as a mixture of updating dynamo tables via both the console and cloudformation in the past caused some inconsistencies between stacks which can have disastrous consequences if confused. Namely, data loss.

LATaylor-guardian:
👍

PR #127, opened 2016-12-22
docker-compose-ify dev bootstrap scripts courtesy of arahayrabedian.

LATaylor-guardian:


LATaylor-guardian:
@arahayrabedian Thank you for this! Finally got around to it with some free time at Christmas. Merging.

arahayrabedian:
awesome! thanks @LATaylor-guardian . If you guys want some tips on docker-compose + dev I wrote [a guide](https://arahayrabedian.github.io/docker-compose-for-dev/) a while back. Don't think it's outdated yet but docker moves fast.

I've also realised I used `restart: always` - you may prefer to play with that setting - it was new to me at the time but it solved the issue of the container dying before cassandra was ready. I've since seen [this breakdown](https://blog.codeship.com/ensuring-containers-are-always-running-with-dockers-restart-policy/) on the options for that. Since we only want it to restart during setup/initialization you can tune that. TL;DR - if you find the kong container coming up after restarts and such and that annoys you, play with this setting, probably change to `restart: on-failures` but I haven't tested that.

PR #128, opened 2017-01-24
uses account id token in cf

tsop14:


LATaylor-guardian:
👍  Be careful to apply the right cloudformation to the right stack! Mixing them up can result in important data being deleted from Dynamo.

PR #129, opened 2017-02-15
Riff raff yaml

tomrf1:
Tested in CODE


riff-raff.yaml
```diff
@@ -0,0 +1,20 @@
...
+      bucket: content-api-dist
+  bonobo-ami-update:
+    type: ami-cloudformation-parameter
+    parameters:
+      amiTags:
+        Recipe: ubuntu-wily-java8
+        AmigoStage: PROD
+      cloudFormationStackName: bonobo
+      prependStackToCloudFormationStackName: false
+      cloudFormationStackByTags: false
```
LATaylor-guardian: Can we add the following please to the `bonobo-ami-update` task: `dependencies: [bonobo]`.

It should be on the same level as parameters, see: https://github.com/guardian/okr-weekly/blob/master/conf/riff-raff.yaml

LATaylor-guardian:
Other than my comment 👍

PR #130, opened 2017-04-10
Maintenance mode

tomrf1:
For use during Kong upgrades. This will reject all requests (besides /healthcheck) with:
`503 - Server is down for maintenance, please try again later`

This is controlled with a config item, `maintenanceMode`, and Bonobo has to be redeployed to switch on/off. This feature will be used so rarely that it's not really worth doing something more sophisticated.


LATaylor-guardian:
👍

PR #131, opened 2017-08-31
Upgrade to Play 2.6

regiskuckaertz:
Almost there.

- [x] make i18n work in tests

### DIFF
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -94,8 +94,8 @@ trait KongComponentImpl extends KongComponent { self: BuiltInComponents with Ahc
   def confString(key: String) = configuration.getOptional[String](key) getOrElse sys.error(s"Missing configuration key: $key")
 
   val kong = {
-    val apiAddress = confString("kong.apiAddress")
-    val apiName = confString("kong.apiName")
+    val apiAddress = confString("kong.new.apiAddress")
+    val apiName = confString("kong.new.apiName")
     new KongClient(wsClient, apiAddress, apiName)
   }
 }
@@ -184,5 +184,5 @@ class AppComponents(context: Context)
   with AssetsComponents
   with ControllersComponent {
 
-  def enableAuth = false
+  def enableAuth = true
 }
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -94,8 +94,8 @@ trait KongComponentImpl extends KongComponent { self: BuiltInComponents with Ahc
   def confString(key: String) = configuration.getOptional[String](key) getOrElse sys.error(s"Missing configuration key: $key")
 
   val kong = {
-    val apiAddress = confString("kong.new.apiAddress")
-    val apiName = confString("kong.new.apiName")
+    val apiAddress = confString("kong.apiAddress")
+    val apiName = confString("kong.apiName")
     new KongClient(wsClient, apiAddress, apiName)
   }
 }
--- a/app/components/AppComponents.scala
+++ b/app/components/AppComponents.scala
@@ -5,9 +5,11 @@ import java.io.FileInputStream
 import akka.stream.Materializer
 import auth.{ Authorisation, DummyAuthorisation, GoogleGroupsAuthorisation }
 import com.amazonaws.regions.Regions
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder
 import com.amazonaws.services.dynamodbv2.document.DynamoDB
-import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceAsyncClient
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceAsyncClientBuilder
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceAsync
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import controllers._
 import com.gu.googleauth.{ GoogleAuthConfig, GoogleServiceAccount }
@@ -39,8 +41,7 @@ trait GoogleAuthComponent { self: BuiltInComponents =>
       redirectUrl = configuration.get[String]("google.redirectUrl"),
       domain = "guardian.co.uk",
       maxAuthAge = Some(Duration.standardDays(90)),
-      enforceValidity = true
-    )
+      enforceValidity = true)
   }
 
 }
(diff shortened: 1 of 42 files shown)


app/components/AppComponents.scala
```diff
@@ -11,13 +12,13 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import controllers._
 import com.gu.googleauth.{ GoogleAuthConfig, GoogleServiceAccount }
 import email.{ AwsEmailClient, MailClient }
+import javax.inject.Inject
```
annebyrne: Why did you decide to go down the DI route? Just for curiousity's sake

app/components/AppComponents.scala
```diff
@@ -105,10 +106,10 @@ trait AwsEmailComponent {
 
 trait AwsEmailComponentImpl extends AwsEmailComponent { self: BuiltInComponents =>
   val awsEmail = {
-    val awsRegion = Regions.fromName(configuration.getString("aws.region") getOrElse "eu-west-1")
+    val awsRegion = Regions.fromName(configuration.getOptional[String]("aws.region") getOrElse "eu-west-1")
```
annebyrne: are we setting this twice in the same way?

app/components/AppComponents.scala
```diff
@@ -147,30 +148,23 @@ trait FiltersComponent extends CSRFComponents { self: BuiltInComponents =>
...
     with KongComponentImpl
     with AwsEmailComponentImpl
     with LabelsComponentImpl
-    with ControllersComponent
-    with FiltersComponent {
-  def enableAuth = true
+    with FiltersComponent
+    with AssetsComponents {
+
+  def appController = new Application(controllerComponents, dynamo, kong, awsEmail, labelsMap, googleAuthConfig, assetsFinder, None)
```
annebyrne: Worth being explicit here about why we're passing a `None`?

app/controllers/Application.scala
```diff
@@ -3,91 +3,94 @@ package controllers
...
 import kong.Kong._
 
-import scala.concurrent.ExecutionContext.Implicits.global
+import javax.inject.Inject
+import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
-class Application(dynamo: DB, kong: Kong, awsEmail: MailClient, labelsMap: Map[String, LabelProperties], val messagesApi: MessagesApi, val authConfig: GoogleAuthConfig, val enableAuth: Boolean) extends Controller
-    with AuthActions
+class Application @Inject() (override val controllerComponents: ControllerComponents, dynamo: DB, kong: Kong, awsEmail: MailClient, labelsMap: Map[String, LabelProperties], val authConfig: GoogleAuthConfig, assetsFinder: AssetsFinder, maybeAuth: Option[ActionBuilder[AuthReq, AnyContent]])(implicit ec: ExecutionContext)
```
annebyrne: As above, curious why you chose DI

app/controllers/AuthActions.scala
```diff
@@ -1,8 +0,0 @@
-package controllers
```
annebyrne: 😁

test/controllers/ApplicationSpec.scala
```diff
@@ -2,24 +2,33 @@ package controllers
...
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class ApplicationSpec extends FlatSpec with Matchers with MockitoSugar with StubBodyParserFactory with StubControllerComponentsFactory {
 
   val mockDynamo = mock[DB]
   val mockKong = mock[Kong]
   val mockEmail = mock[MailClient]
-  val messagesApi = new DefaultMessagesApi(Environment.simple(), Configuration.reference, new DefaultLangs(Configuration.reference))
+  val mockActionBuilder = new AuthenticatedBuilder[UserIdentity](
+    userinfo = _ => Some(UserIdentity("", "", "First", "Last", Long.MaxValue, None)),
```
annebyrne: :+1:

app/components/AppComponents.scala
mchv: this should be removed we don't want to inject anything

app/controllers/Application.scala
mchv: that should be reomved, we don't use DI

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.13
+sbt.version=0.13.15
```
mchv: I think `0.13.16` is the latest before `1.0.0`

app/components/AppComponents.scala
annebyrne: That was my instinct but can you explain why we tend not to go down a DI route

LATaylor-guardian: Magic ✨

LATaylor-guardian: It's still using DI just compile time DI i.e passing dependencies via the constructor of a class.

LATaylor-guardian: We prefer compile time DI over runtime DI because there's less magic. I.e it won't compile if we have a problem rather than blowing up at runtime. It also keeps things simple - something needs a dependency so we explicitly give it rather than 'wiring' it in magically.

annebyrne: Thank you!! @LATaylor-guardian

mchv: @annebyrne I have tried to explain why dependency injection is an anti-pattern in the [guardian recommendation for scala](https://github.com/guardian/recommendations/blob/master/scala.md)

regiskuckaertz: Ha! Thanks for the reminder, yes I meant to use regular arguments.

app/components/AppComponents.scala
regiskuckaertz: looks like it yeah

app/components/AppComponents.scala
regiskuckaertz: Good call, will comment on that!

app/controllers/Application.scala
regiskuckaertz: will remove.

test/controllers/ApplicationSpec.scala
```diff
@@ -66,21 +75,21 @@ class ApplicationSpec extends FlatSpec with Matchers with MockitoSugar {
...
-    val application = new Application(dynamo, mockKong, mockEmail, Map.empty, messagesApi, null, false)
+    val application = new Application(mockCC, dynamo, mockKong, mockEmail, Map.empty, null, mockAssetsFinder, Some(mockActionBuilder))
     val result: Future[Result] = application.showKeys(List.empty, "next", None).apply(FakeRequest())
     contentAsString(result) should include("my-new-key")
   }
 
   "brokenForm" should "check form validation works" in {
     val myRequest: (String, String) = ("name", "")
-    val application = new Application(mockDynamo, mockKong, mockEmail, Map.empty, messagesApi, null, false)
+    val application = new Application(mockCC, mockDynamo, mockKong, mockEmail, Map.empty, null, mockAssetsFinder, Some(mockActionBuilder))
```
regiskuckaertz: If someone could help me figure out why Play does not seem to find the default resource dictionary. It outputs `error.required` instead of `This field is required`.

regiskuckaertz: Fixed it by passing a mock `MessagesApi` object in the spec.

app/controllers/csrf/CSRFSupport.scala
```diff
@@ -9,9 +9,8 @@ object CSRFSupport {
   val config = CSRFConfig()
 
   def getToken(request: RequestHeader): Option[String] = {
-    request.tags.get(Token.RequestTag)
-      // Check cookie if cookie name is defined
-      .orElse(config.cookieName.flatMap(n => request.cookies.get(n).map(_.value)))
+    // Check cookie if cookie name is defined
```
regiskuckaertz: This change I'm not 100% sure, but `RequestTag` has been removed from `Token` and the only occurence left is in `CORSFilter`, which is not `CSRF`.

annebyrne: 🤔  Guess we'll just have to see in integration testing whether this works!

regiskuckaertz:
Tested on CODE, everything looks fine.

tomrf1:
👍 looks good to me

PR #132, opened 2017-10-03
Fix broken registration

mchv:
The [migration to play 2.6](https://github.com/guardian/bonobo/pull/131) break the CSRF token to be populated correcly

@tackley


tackley:
😃

mchv:
tested on `PROD` works now as expected!

PR #135, opened 2018-03-15
Add the ability to delete keys

regiskuckaertz:
This is a first stab at adding the ability to delete API keys. Fortunately the deletion logic in Kong is already there, I just had to add the plumbing.

There is a heavy REST flavour to any Play app so I went all in: I added a DELETE route. 

Sadly, this automatically requires using ajax on the client side. Why `<form action="..." method="delete">` is not allowed is completely beyond me 😩 As a consequence, what happens after deletion must then be handled entirely in JavaScript. The current architecture does not leave a lot of room for callbacks, so I just added some redirect logic for now.

I'm already regretting this decision and wonder if instead I should just add a good old POST route with a path like `/key/:id/delete` and call it a day.

### DIFF
--- a/conf/routes
+++ b/conf/routes
@@ -12,7 +12,7 @@ GET         /key/create/:userId         controllers.Application.createKeyPage(us
 POST        /key/create/:userId         controllers.Application.createKey(userId: String)
 GET         /key/:id/edit               controllers.Application.editKeyPage(id: String)
 POST        /key/:id/edit               controllers.Application.editKey(id: String)
-DELETE      /key/:id                    controllers.Application.deleteKey(id: String)
+DELETE      /key/:id/edit               controllers.Application.deleteKey(id: String)
 
 GET         /user/create                controllers.Application.createUserPage
 POST        /user/create                controllers.Application.createUser
--- a/conf/routes
+++ b/conf/routes
@@ -12,6 +12,7 @@ GET         /key/create/:userId         controllers.Application.createKeyPage(us
 POST        /key/create/:userId         controllers.Application.createKey(userId: String)
 GET         /key/:id/edit               controllers.Application.editKeyPage(id: String)
 POST        /key/:id/edit               controllers.Application.editKey(id: String)
+DELETE      /key/:id                    controllers.Application.deleteKey(id: String)
 
 GET         /user/create                controllers.Application.createUserPage
 POST        /user/create                controllers.Application.createUser
--- a/public/javascripts/deleteKey.js
+++ b/public/javascripts/deleteKey.js
@@ -1,0 +1,13 @@
+Array.from(document.querySelectorAll('.js-delete-key')).forEach(btn => {
+    btn.onclick = e => {
+        e.stopPropagation();
+        let req = new XMLHttpRequest();
+        let url = btn.getAttribute('data-url');
+        let red = btn.getAttribute('data-redirect-url');
+        req.onload = req.onerror = () => {
+          window.location.href = red;
+        };        
+        req.open('DELETE', url, true);
+        req.send();
+    };
+});
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -222,8 +222,11 @@ class Application(
     editKeyForm.bindFromRequest.fold[Future[Result]](handleInvalidForm, handleValidForm)
   }
 
-  def deleteKey(keyId: String) = maybeAuth.async { implicit request => 
-    
+  def deleteKey(keyId: String) = maybeAuth.async { implicit request =>
+    dynamo.getKeyWithValue(keyId) match {
+      case Some(key) => logic.deleteKey(key)
+      case None => Future.successful(NotFound)
+    }    
   }
 
   def getEmails(tier: String, status: String) = maybeAuth { implicit request =>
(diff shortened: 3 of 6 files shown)


hosamaly:
> Why `<form action="..." method="delete">` is not allowed is completely beyond me

What should an HTML-only browser do when it receives the response to this request? As described in [a discussion in W3C](https://www.w3.org/Bugs/Public/show_bug.cgi?id=10671#c0):
> server implementations usually respond with 200 and a minimal response body ("deleted") or 204 (no content). So it's not clear how this can be used in a web application.

In the discussion link above, you can see examples of how different frameworks handle this. For example, Ruby on Rails generates an extra form attribute and then processes the request on the server side before routing. It's possible to mimic this using a Play filter but [there are security risks](https://github.com/playframework/playframework/issues/2063).

An interesting summary of the W3C discussions can be found [here](https://softwareengineering.stackexchange.com/a/211790/7671).

conf/routes
```diff
@@ -12,6 +12,7 @@ GET         /key/create/:userId         controllers.Application.createKeyPage(us
 POST        /key/create/:userId         controllers.Application.createKey(userId: String)
 GET         /key/:id/edit               controllers.Application.editKeyPage(id: String)
 POST        /key/:id/edit               controllers.Application.editKey(id: String)
+DELETE      /key/:id/edit               controllers.Application.deleteKey(id: String)
```
hosamaly: I think that a pure REST approach would be `DELETE /key/:id`. Although, as mentioned in another conversation, I think that you're right to switch to good old `POST`.

mchv: 👍 for POST

regiskuckaertz:
Well I have to say, `server implementations usually respond with 200 and a minimal response body` is really no excuse for _thinking_. I find it baffling that on the back of that petty excuse they removed an essential feature of the HTTP model. A DELETE request has all the information it needs for the middleware to make a reasoned decision about what the response should be, just like a POST request or even a GET.

Anyway, moving to POST then 😡

public/javascripts/deleteKey.js
```diff
@@ -0,0 +1,13 @@
+Array.from(document.querySelectorAll('.js-delete-key')).forEach(btn => {
+    btn.onclick = e => {
```
hosamaly: It would be nice to ask the user for confirmation. It can be as simple as a call to `confirm`. (Or, if UX is valuable here, perhaps expand the button on the first click and act upon the second.)

app/store/Dynamo.scala
```diff
@@ -173,6 +175,11 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
     Logger.info(s"DynamoDB: Key ${kongKey.key} has been updated")
   }
 
+  def deleteKey(kongKey: KongKey): Unit = {
+    KongTable.deleteItem(
+      new PrimaryKey("hashkey", "hashkey", "rangekey", kongKey.rangeKey))
```
hosamaly: This expression is repeated five times, and some constants are repeated >15 times (e.g. `"hashkey"`). Do you have an appetite to deduplicate them?

regiskuckaertz:
Tested in CODE

screen shot 2018-03-19 at 15 46 31
screen shot 2018-03-19 at 15 46 41

app/store/Dynamo.scala
regiskuckaertz: I think this app needs a refactoring, some things I found mind-boggling, like the fact the value in the `hashkey` column is always the string `"hashkey"` 🤔

cloudformation-CODE.yaml
```diff
@@ -152,7 +152,7 @@ Resources:
         Fn::Base64:
           !Sub |
             #!/usr/bin/env bash
-            wget -NP /home/ubuntu/.ssh --no-check-certificate https://s3-eu-west-1.amazonaws.com/content-api-dist/authorized_keys
+            curl https://s3-eu-west-1.amazonaws.com/content-api-dist/authorized_keys >> /home/ubuntu/.ssh/authorized_keys
```
hosamaly: Whenever I use `curl` in a script, I find it useful to pass the following options to avoid unexpected behaviour:
> [**-L, --location**](https://curl.haxx.se/docs/manpage.html#-L)
> (HTTP) If the server reports that the requested page has moved to a different location (indicated with a Location: header and a 3XX response code), this option will make curl redo the request on the new place.

> [**-f, --fail**](https://curl.haxx.se/docs/manpage.html#-f)
> (HTTP) Fail silently (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting that and return error 22.
> This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).

> [**-S, --show-error**](https://curl.haxx.se/docs/manpage.html#-S)
> When used with -s, --silent, it makes curl show an error message if it fails.

app/store/Dynamo.scala
tomrf1: 😆

tomrf1:
👍 nice!

PR #138, opened 2018-05-09
Add the ability for users to extend and/or remove their own keys

regiskuckaertz:
The GDPR requires us to make sure we get rid of API keys that are no longer being used, i.e. that have not been used in the last 30 months (gasp). Sadly, the key management feature in Kong does not allow us to poll for a key and detect whether that is the case or not. So instead we've got a workflow that will work as is:

1 - Poll the database daily for keys that were created 30 months ago, or more. Send an email to their respective owner asking them to review their keys (this is done by [a separate lambda](https://github.com/guardian/gibbons))
2 - The email contains a link to a page where the user can take appropriate action (extend or delete)
3 - If the user does not take action within two weeks of the email being sent, the keys will be automatically deleted (also done in the same lambda)

This PR addresses 2. Nothing very complex here... except for one thing: authentication. We don't want anyone to do anything unless expressively requested.

There's a couple of things that remain to be done:

- [x] test key deletion in CODE
-  ~~highlight expired keys (or those which are about to expire)~~
- [x] add super fancy stuff to the public URLs to prevent tampering with our keys
- [x] add delete user from Kong and Bonobo repositories
- [x] write tests

@tomrf1 is of the opinion that we shouldn't even show users their own keys, just two buttons ("extend my keys" and "delete my keys"), on the principle that the less data we show, the better. I'm happy to go with that 😄


regiskuckaertz:
The last change, as a matter of precaution: instead of sending the user's ID in clear via a URL, it'll be hashed and stored in the DB. Also, when fetching the user, we make sure the request happens within two weeks of the email being sent. This is not a perfect solution, but it will mitigate brute-force attacks, should that ever happen.

app/components/AppComponents.scala
```diff
@@ -129,6 +130,21 @@ trait LabelsComponentImpl extends LabelsComponent with DynamoComponent {
   }
 }
 
+trait HashComponent {
+  val digest = MessageDigest.getInstance("MD5")
+
+  def md5(str: String): String = {
+    val hash = str + salt
+    digest.digest(hash.getBytes).map("%02X".format(_)).mkString
+  }
```
mchv: `MD5`  is not considered secured and should not be used.
Please switch to `SHA3-512`.
You can use it via [BouncyCastly](http://www.bouncycastle.org/) library: `org.bouncycastle.jcajce.provider.digest.SHA3.DigestSHA3`

mchv: See https://en.wikipedia.org/wiki/Template:Comparison_of_SHA_functions for comparison of hashing functions

hosamaly: I'm not sure what's the benefit of hashing here. If the request is authenticated and authorised (i.e. a regular user can only delete his/her key), then what's the benefit of hashing?
(Especially knowing that key ids are UUIDs, not sequential.)

app/controllers/DeveloperForm.scala
```diff
@@ -44,6 +44,32 @@ class DeveloperForm(override val controllerComponents: ControllerComponents, dyn
...
   }
 
+  def deleteKeys(id: String, hash: String) = Action.async { implicit request =>
+    if (md5(id) != hash)
+      Future.successful(Forbidden)
+    else
+      logic.deleteKeys(id).recover {
+        case err =>
+          awsEmail.sendEmailDeletionFailed(id, err)
+          ()
```
hosamaly: Should we return an appropriate error message here?

app/email/AwsEmailClient.scala
```diff
@@ -99,4 +103,38 @@ class AwsEmailClient(amazonMailClient: AmazonSimpleEmailServiceAsync, fromAddres
...
+      s"""Damn Daniel.
+         |
+         |Something went wrong when trying to delete assets for user
+         |
+         |     $userId
+         |
+         |As a result, all their assets are still there. Sorry but you'll have to investigate.
+         |Here's all I've got:
+         |
+         | $err
```
hosamaly: We may want to include the messages of suppressed exceptions.

app/kong/Kong.scala
```diff
@@ -135,6 +137,15 @@ class KongClient(ws: WSClient, serverUrl: String, apiName: String) extends Kong
       }
   }
 
+  def deleteConsumer(consumerId: String): Future[Happy.type] = {
+    ws.url(s"$serverUrl/consumers/$consumerId").delete().flatMap { response =>
+      response.status match {
+        case 204 => success(s"Kong: Successfully deleted consumer $consumerId", Happy)
+        case _ => genericFail(s"Kong: Failed to delete consumer $consumerId. Response ${response.status} ${response.statusText}")
```
hosamaly: Can we include `response.body`, or log it if its contents may be sensitive?

app/logic/DeveloperFormLogic.scala
```diff
@@ -41,4 +42,30 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
...
     }
   }
+
+  def deleteKeys(id: String): Future[Unit] = {
+    dynamo.getUserWithId(id).fold(userNotFound(id): Future[Unit]) { user =>
+      val keys = dynamo.getKeysWithUserId(user.bonoboId)
+      Future.traverse(keys) { key =>
+        for {
+          _ <- kong.deleteKey(key.kongConsumerId)
+          _ <- kong.deleteConsumer(key.kongConsumerId)
```
hosamaly: Do both `deleteKey` and `deleteConsumer` take the same `kongConsumerId`? I would've expected `deleteKey` to accept `key.bonoboId`...

app/logic/DeveloperFormLogic.scala
```diff
@@ -41,4 +42,30 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
       }
     }
   }
+
+  def deleteKeys(id: String): Future[Unit] = {
```
hosamaly: Since this method deletes the user as well as its keys, I suggest renaming it to `deleteUser`. Until then, we should rename the parameter to `userId`.

app/logic/DeveloperFormLogic.scala
```diff
@@ -41,4 +42,30 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
...
+        dynamo.deleteUser(user)
+      }
+    }
+  }
+
+  def extendKeys(id: String): Future[Unit] = {
+    dynamo.getUserWithId(id).fold(userNotFound(id): Future[Unit]) { user =>
+      val keys = dynamo.getKeysWithUserId(user.bonoboId)
+      val now = Some(DateTime.now())
+      Future.traverse(keys)(key => Future.successful(dynamo.updateKey(key.copy(extendedAt = now)))).map { _ => () }
```
hosamaly: In this expression, if updating a single key fails then we won't try to update the rest.
If desired, we can continue to try and update the remaining keys, even if some of them fail, by changing the call to `Future.successful` to `Future.apply`.
```scala
Future.traverse { key => Future { dynamo.updateKey(key.copy(extendedAt = now)) } }
```

app/logic/DeveloperFormLogic.scala
```diff
@@ -41,4 +42,30 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
...
+        } yield {
+          dynamo.deleteKey(key)
+        }
+      }.map { _ =>
+        dynamo.deleteUser(user)
+      }
+    }
+  }
+
+  def extendKeys(id: String): Future[Unit] = {
```
hosamaly: Is there a particular reason to specify the return type as `Future[Unit]`?

My personal preference is to return `Future[Seq[KongKey]]`. But, if this is not desired, we can specify it as `Future[_]` and avoid the call to `.map { _ => () }`.

app/logic/DeveloperFormLogic.scala
```diff
@@ -41,4 +42,30 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
...
+    dynamo.getUserWithId(id).fold(userNotFound(id): Future[Unit]) { user =>
+      val keys = dynamo.getKeysWithUserId(user.bonoboId)
+      val now = Some(DateTime.now())
+      Future.traverse(keys)(key => Future.successful(dynamo.updateKey(key.copy(extendedAt = now)))).map { _ => () }
+    }
+  }
+
+  private def userNotFound(id: String) = Future.failed(new Throwable(s"User $id not found"))
```
hosamaly: May I suggest using `NoSuchElementException`, `IllegalArgumentException`, or `RuntimeException` instead of `Throwable`?

app/logic/DeveloperFormLogic.scala
```diff
@@ -41,4 +42,30 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
       }
     }
   }
+
+  def deleteKeys(id: String): Future[Unit] = {
+    dynamo.getUserWithId(id).fold(userNotFound(id): Future[Unit]) { user =>
```
hosamaly: The call to `dynamo.getUserWithId` [may throw an exception](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/AmazonDynamoDB.html#query-com.amazonaws.services.dynamodbv2.model.QueryRequest-). This will bubble up to the caller of this method.

Since this method returns a `Future`, I think that it's better to encapsulate any exceptions thrown inside it into a `Future`.

app/views/developerDeleteComplete.scala.html
```diff
@@ -0,0 +1,14 @@
+@import controllers.AssetsFinder
+@(assets: AssetsFinder, success: Option[String] = None, error: Option[String] = None)
+
+@mainOpenForm(assets, "Show key", success, error) {
+    <div>
+        <div class="row">
+            <div class="col-md-8 col-md-offset-2 column">
+                <p>Thank you!</p>
+                <p>You have successfully deleted all your keys.</p>
+                <p>If you ever want to get a new one, head over <a href="https://bonobo.capi.gutools.co.uk/register/developer">the registration form</a>.</p>
```
hosamaly: I think that this should say "head over _to_ ..."

integration-test/scala/integration/IntegrationSpecBase.scala
```diff
@@ -35,6 +35,10 @@ class FakeEmailClient extends MailClient {
   def sendEmailCommercialRequestToUser(toEmail: String): Future[SendEmailResult] = Future.failed(new Exception("Error when sending emails for commercial request to user"))
 
   def sendEmailNewKey(toEmail: String, key: String): Future[SendEmailResult] = Future.failed(new Exception("Error when sending emails for new key"))
+
+  def sendEmailDeletionFailed(userId: String, err: Throwable): Future[SendEmailResult] = Future.failed(new Exception("Error when sending emails for a failed deletion"))
+
+  def sendEmailExtensionFailed(userId: String, err: Throwable): Future[SendEmailResult] = Future.failed(new Exception("Error when sending emails for a failed deletion"))
```
hosamaly: The message should say "for a failed _extension_".

app/components/AppComponents.scala
regiskuckaertz: The routes are public and not authenticated. Better safe than sorry, we can get rid of that part should it prove irrelevant.

app/controllers/DeveloperForm.scala
regiskuckaertz: I'm on the fence. The time-sensitive nature of the key cleanup process means if any error occurs, we should react immediately. _If_ an error message was displayed to the user, say "Something's wrong, please retry later", there is a likelihood they will forget and lose their key.

app/email/AwsEmailClient.scala
```diff
@@ -99,4 +103,38 @@ class AwsEmailClient(amazonMailClient: AmazonSimpleEmailServiceAsync, fromAddres
...
+         |
+         |Good luck.""".stripMargin
+    sendEmail(fromAddress, "Keys deletion failed", message)
```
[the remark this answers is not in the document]
regiskuckaertz: Good point 👍

app/logic/DeveloperFormLogic.scala
regiskuckaertz: It is the same ID indeed, `kongConsumerId` is a "foreign key" into the Kong database, where we store one API key per consumer (Kong has no knowledge of `bonoboId`)

app/logic/DeveloperFormLogic.scala
regiskuckaertz: Another use of `_` I didn't know of 👍

app/logic/DeveloperFormLogic.scala
regiskuckaertz: Thanks 👍

mchv:
Regarding authentication the way to do it is to use a token generated server side from the lambda, valid for only few days, stored in dynamo, and check when a user hit that `url` that the token is exist. It is very similar to a `CRSF`.

@tomrf1 is right that we should not expose user id, so we should keep the hash in the url to identify the user, but not considerate a legitimate request without a token stored server-side and only valid for **once** for a limited period of time.

mchv:
Without a `token` you can easily delete keys with a basic bruteforce attack or delete a specific key if you have access to the secret url (even if leaked long time later the email has been sent)

regiskuckaertz:
@mchv at first I was hashing the user ID too, but I thought a hash with limited lifetime was enough (since if the URL is ever shared, that it contains the user ID or not does not change anything, as the hash _is_ the user's ID by extension).

regiskuckaertz:
Ok so, lifetime of the token.

- on delete, since we delete everything, nothing needs to be done
- on extend, the time used to compute the hash is deleted from the DB, which makes the hash unusable in the future.

Now trying to make the SHA3 thing work. It throws `NullPointerException` when I create an instance 😭

app/controllers/DeveloperForm.scala
```diff
@@ -44,30 +45,45 @@ class DeveloperForm(override val controllerComponents: ControllerComponents, dyn
...
-      logic.extendKeys(id).recover {
-        case err =>
-          awsEmail.sendEmailExtensionFailed(id, err)
-          ()
-      }.map {
-        _ => Ok(views.html.developerExtendComplete(assetsFinder))
-      }
+  def extendUser(userId: String, hash: String) = Action.async { implicit request =>
+    Future { dynamo.getUserWithId(userId) } flatMap {
+      case None => Future.successful(NoContent)
```
hosamaly: Was the change from `Forbidden` to `NoContent` intentional? If so, may I ask why?

regiskuckaertz: `Forbidden` is still there, but only if the hash is wrong.

hosamaly: But we're telling the client that the operation succeeded even though we couldn't find the user. Shouldn't we inform them that it failed?
I understand that 404 may not be desirable for security purposes, so 403 should tell the client that something is wrong.

app/controllers/DeveloperForm.scala
```diff
@@ -44,30 +45,45 @@ class DeveloperForm(override val controllerComponents: ControllerComponents, dyn
...
+          }.map {
+            _ => Ok(views.html.developerExtendComplete(assetsFinder))
+          }
+        else
+          Future.successful(Forbidden)
+    }
+  }
+
+  def validate(userId: String, hash: String)(time: Long): Boolean = {
+    val twoWeeksAgo = DateTime.now.minusWeeks(2).getMillis
```
hosamaly: We should extract this to a configuration property. It should be used to calculate the time that should have passed since the last time the key was extended (e.g. 30 months - 2 weeks).

app/logic/DeveloperFormLogic.scala
```diff
@@ -43,29 +43,29 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
...
-        for {
-          _ <- kong.deleteKey(key.kongConsumerId)
-          _ <- kong.deleteConsumer(key.kongConsumerId)
-        } yield {
-          dynamo.deleteKey(key)
-        }
-      }.map { _ =>
-        dynamo.deleteUser(user)
+  def deleteUser(user: BonoboUser): Future[_] = {
+    val keys = dynamo.getKeysWithUserId(user.bonoboId)
```
hosamaly: This call can still throw an exception. To avoid those bubbling up from this method, we can wrap the call in a `Future` (like you did in `DeveloperForm`).
```scala
Future { dynamo.getKeysWithUserId(user.bonoboId) }.flatMap { keys => Future.traverse(keys) { key => ...
```

app/logic/DeveloperFormLogic.scala
```diff
@@ -43,29 +43,29 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
...
+    val newUser = user.copy(additionalInfo = user.additionalInfo.copy(extendedAt = now))
+    Future {
+      dynamo.saveUser(newUser)
+      newUser
     }
   }
 
-  private def userNotFound(id: String) = Future.failed(new Throwable(s"User $id not found"))
+  def invalidateHash(user: BonoboUser): Future[_] =
+    Future { dynamo.saveUser(user.copy(additionalInfo = user.additionalInfo.copy(remindedAt = None))) }
```
hosamaly: This logic can (and I think should) be inlined into `extendUser`.

app/controllers/DeveloperForm.scala
```diff
@@ -45,44 +45,50 @@ class DeveloperForm(override val controllerComponents: ControllerComponents, dyn
     createKeyForm.bindFromRequest.fold[Future[Result]](handleInvalidForm, handleValidForm)
   }
 
+  /**
+   * Extends the lifetime of a user's account and keys. This action is subsequent of
```
hosamaly: `s/Extends/Deletes/` :)

regiskuckaertz:
Tested in CODE ✅

PR #140, opened 2018-07-05
Update email address

mchv:
Not sure why the `.com` does not work, but not a big deal


mchv:
@regiskuckaertz @LATaylor-guardian

PR #141, opened 2018-08-07
Migrate to Teamcity

tomrf1:
also switch to sbt-assembly and sbt v1


tomrf1:
please ignore the reported circleci failure for now, I'll remove it from there after

tomrf1:
@regiskuckaertz good point! I'll switch

PR #142, opened 2018-10-12
Use 'local' rate-limiting policy

tomrf1:
Brings bonobo in line with [this change](https://github.com/guardian/content-api/pull/2184) to the kong rate-limiting plugin.
Drastically reduces DB connections (and instance size), at the cost of less reliable rate-limiting


LATaylor-guardian:
👍

PR #146, opened 2019-10-07
Cleanup build, restore ZIP archive

regiskuckaertz:
This restores the previous build. I don't know why I couldn't see the generated ZIP in the s3 bucket, it is now there and I've tested on CODE.


regiskuckaertz:
No idea 🙈

PR #147, opened 2020-11-27
Script to look at developer tier keys

Mullefa:
## What does this change?
See newly added README for context.

Regarding the new index `keyValue-index-all-attributes` that was created. This was due to the method `getKeyWithValue()` performing a full table scan. In a subsequent PR, I will update the application to use this new index. Note I haven't cloud formed the new index since the resources associated with the Bonobo cloudformation template(s) have drifted; I see resolving this as a separate task.

### DIFF
--- a/dev/developer-api-keys/README.md
+++ b/dev/developer-api-keys/README.md
@@ -1,0 +1,18 @@
+# Developer API keys
+
+Script to understand which API keys currently being used are issued under Developer tier. This is relevant since Gibbons 
+deletes keys in the Developer tier, so we want to eyeball the generated list for any that might be used in PROD.
+
+## API key usage file
+
+The script takes as an argument the location of a CSV file with API key usage. To generate this file:
+ElasticSearch SQL can be used to perform a count of requests by API key (see below). The query can be conveniently
+executed in the Dev Tools section of Kibana. The results can then be exported to a temporary local file for use by the 
+script, but NOT checked into version control!
+
+```
+GET _sql?format=csv
+{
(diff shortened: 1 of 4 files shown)


JustinPinner (approved):
Looks like it'll do what it's supposed to. The concept of passing a reference to a CSV file produced from a Kibana query seems quite odd and outside of Bonobo's normal behaviour, but the code appears sound I think.

dev/developer-api-keys/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Script to understand which API keys currently being used are issued under Developer tier. This is relevant since Gibbons 
+deletes keys in the Developer tier, so we want to eyeball the generated list for any that might be used in PROD.
+
+## API key usage file
+
+The script takes as an argument the location of a CSV file with API key usage. To generate this file:
+ElasticSearch SQL can be used to perform a count of requests by API key (see below). The query can be conveniently
+executed in the Dev Tools section of Kibana. The results can then be exported to a temporary local file for use by the 
+script, but NOT checked into version control!
```
juanvaccari, line 11: 👍

dev/developer-api-keys/src/main/scala/com/gu/bonobo/Main.scala
```diff
@@ -0,0 +1,126 @@
...
+        .action((labelsTable, config) => config.copy(labelsTable = labelsTable)),
+      opt[String]("api-key-usage")
+        .text("see README on how this file can be generated; never check this file into version control")
+        .action((apiKeyUsageFile, config) => config.copy(apiKeyUsageFile = apiKeyUsageFile))
+        .required(),
+      opt[Int]("usage-count")
+        .text("not interested in API keys with usage count < this number")
+        .action((usageCount, config) => config.copy(usageCount = usageCount))
+    )
+  }
```
juanvaccari, line 59: 👍  very nice!

Mullefa:
Hey Justin - yeah, using a CSV perhaps isn't ideal. For context, the primary concern was to avoid (the time consuming part of) looking up each key in the list manually. Generating the CSV in contrast is quick to do (especially compared with the development time to get this data programatically). Note I wouldn't be too concerned about "Bonobo's normal behaviour" since this isn't a change to the application; rather a separate project which is bootstrapping Bonobo's functionality.

juanvaccari (approved):

PR #150, opened 2020-12-21
Single cloud formation template for CODE and PROD stages

Mullefa:
## What does this change?

To facilitate the work of (c.f.) sending Bonobo logs to ELK. 

Amend PROD cloud formation template so it can be used for CODE too; this removes the need for stack updates to require updates to two cloud formation templates.


davidfurey (approved):
lgtm

juanvaccari (approved):

cloudformation.yaml
```diff
@@ -231,6 +243,7 @@ Resources:
         !GetAtt BonoboSecurityGroup.GroupId
   UsersTable:
     Type: AWS::DynamoDB::Table
+    Condition: ShouldCloudFormDynamoDBTables
```
tonytw1, line 246: Any chance of Cloud Formation taking this as an instruction to drop existing tables in CODE.
Is the exisitng prod environment actually tagged 'PROD'.

Mullefa: Good catch, but don't think so. Those tables are completely unrelated to the CODE stack. Although when applying the stack change, I'll look at the change set to make sure.

Mullefa:
I've just updated the CODE stack (the CODE and PROD templates had a difference of a new line in the `UserData` script which meant a valid change set was created).

When trying to update the PROD stack I received this message:

Screenshot 2020-12-23 at 08 48 24

This is good news. In particular it means the condition introduced is not creating a change set which removes the PROD tables!

PR #151, opened 2020-12-23
Send logs to central ELK

Mullefa:
## What does this change?
Sends Bonobo logs to ELK stack. 

I am currently trying to resolve a `_grokparsefailure`. I have narrowed this down to be an issue with the `{JAVACLASS:class_name}`. This means some meta data is not added to application log events. Otherwise, working fine.


JustinPinner (approved):
LGTM 👍  Good to know there's an instance type change prior to the upcoming reservation round too! Huge 💰 saved 😆

PR #152, opened 2021-01-25
Local Kong setup refresh

tonytw1:
## What does this change?

Updates the local Kong setup to match the versions currently in live (0.14 Kong / 9.5 PostgreSQL).

Updates the Dockers usages to match the how Docker currently works on a 2021 dev laptop (ie. we don't need virtualbox).

Verified locally;  Kong pitches up on http://localhost:8000 and 8001.

### DIFF
--- a/scripts/configure-docker.sh
+++ b/scripts/configure-docker.sh
@@ -1,5 +1,4 @@
 #! /bin/bash -e
-pwd
 
 echo Starting Kong Docker images ...
 docker-compose up -d
--- a/scripts/configure-docker.sh
+++ b/scripts/configure-docker.sh
@@ -1,14 +1,22 @@
-#! /bin/bash
+#! /bin/bash -e
+pwd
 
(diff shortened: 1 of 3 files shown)


scripts/configure-docker.sh
```diff
@@ -1,14 +1,22 @@
...
+echo Starting Kong Docker images ...
 docker-compose up -d
 
+echo Waiting for Kong
+wget -O - http://localhost:8001/apis --retry-connrefused --no-verbose
+echo Kong looks responsive
+
 echo Adding API ...
-wget -O - http://${container_host}:8001/apis --post-data 'name=internal&request_host=foo.com&upstream_url=http://example.com' --retry-connrefused --no-verbose
+curl -i -X POST \
```
tonytw1, line 12: Http 400'ed out of the box.
Ported to curl to match the Kong docs.

Mullefa (approved):

PR #153, opened 2021-01-29
Key creation dates fix

tonytw1:
## What does this change?

Fixes incorrect 1970 creation dates for newly created keys.

Kong API 0.14 and high renders date times as epoch seconds. Older versions of the Kong API seem to have used milliseconds.

Screenshot 2021-01-28 at 16 12 34

An old version of Kong (0.9) returning milliseconds. 
We once used Kong 0.5 and bonobo was probably coded against that.

Screenshot 2021-01-28 at 19 58 59

### DIFF
--- a/app/kong/Kong.scala
+++ b/app/kong/Kong.scala
@@ -46,6 +46,8 @@ object Kong {
   }
 
   def consumerCreationResponseFor(consumer: KongCreateConsumerResponse, key: String): ConsumerCreationResult = {
+    // Kong API 0.14 (and higher) renders data as 10 digit seconds since 1970.
+    // Previous Kong API versions (such as 0.9) used a 14 digit milliseconds since 1970 format
     ConsumerCreationResult(consumer.id, new DateTime(new Instant(consumer.created_at * 1000), DateTimeZone.UTC), key)
   }
 }
--- a/app/kong/Kong.scala
+++ b/app/kong/Kong.scala
@@ -73,10 +73,14 @@ class KongClient(ws: WSClient, serverUrl: String, apiName: String) extends Kong 
 
   def createConsumerAndKey(tier: Tier, rateLimit: RateLimits, key: Option[String]): Future[ConsumerCreationResult] = {
     for {
-      consumer <- createConsumer(tier)
-      _ <- setRateLimit(consumer.id, rateLimit)
-      key <- createKey(consumer.id, key)
-    } yield ConsumerCreationResult(consumer.id, new DateTime(consumer.created_at), key)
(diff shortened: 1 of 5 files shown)


app/kong/Kong.scala
```diff
@@ -46,6 +44,10 @@ object Kong {
   object KongPluginConfig {
     implicit val pluginsRead = Json.reads[KongPluginConfig]
   }
+
+  def consumerCreationResponseFor(consumer: KongCreateConsumerResponse, key: String): ConsumerCreationResult = {
+    ConsumerCreationResult(consumer.id, new DateTime(new Instant(consumer.created_at * 1000), DateTimeZone.UTC), key)
```
JustinPinner, line 49: Can we add a comment here about why we're having to * 1000? Because future me will be puzzled, I just know it 😁

scripts/backfill-created-dates.py
```diff
@@ -0,0 +1,50 @@
...
+
+# Get the service resource.
+dynamodb = capi.resource('dynamodb')
+
+
+table = dynamodb.Table('bonobo-CODE-keys')
+print(table)
+
+# A sensible epoch mills datetime will not be less than this
+cutoffDate = 1000000000000
```
JustinPinner, line 20: I found this to be Sunday, September 9, 2001 1:46:40 AM, which sounds very reasonable!

scripts/backfill-created-dates.py
```diff
@@ -0,0 +1,50 @@
...
+
+print(f"Found {response['Count']} / scanned {response['ScannedCount']}")
+
+items = response['Items']
+for item in items:
+    keyValue = item['keyValue']
+    createdAt = item['createdAt']
+    if createdAt < cutoffDate:
+        updatedCreatedAt = createdAt * 1000
+        print(f"Item {keyValue} has old createdAt {createdAt} and will updated to {updatedCreatedAt}")
```
JustinPinner, line 35: Can we capture these outputs in case there are any weird things to investigate afterwards?

JustinPinner (approved):
I thinks this all looks good (added a couple of non-blocking questions). Would be very happy for another opinion too though. In the meantime, I'll go ahead and approve 👍

scripts/backfill-created-dates.py
tonytw1: Yep - that's fair.

app/kong/Kong.scala
tonytw1: Yep - brought the comment across from the test.

PR #154, opened 2021-02-01
Migrate local Kong from deprecated /apis to /services and /routes.

tonytw1:
## What does this change?

Migrate CAPI Kong from deprecated API concept to sevices and routes.
Migrate existing keys and rate limits.

Kong 0.13 deprecated the API concept splitting it into Services with Routes.
https://github.com/Kong/kong/blob/master/CHANGELOG.md#0130---20180322

To ever progress into the 1.0 series and beyond we need to migrate our deprecated /apis/internal to a service and route.

Kong specfically askes us todo this ourselves:
https://github.com/Kong/kong/blob/master/UPGRADE.md#upgrade-from-0x-to-10x

Screenshot 2021-02-04 at 12 35 39

This PR does this by:

1) Creating a new Service to represent back end CAPI.
2) Setting up the keyauth plugin on the new service (/service/internal)
3) Iterating through the rate limit plugin configurations attached to /api/internal and copying them to /service/internal
4) Creating a new route to send requests to the content api hostname through the new service.

The old /apis/internal can then be deleted.

Bonobo is updated to setup rate limits on the service rather than the old api endpoints,

## Why are you even touching this?

Our version of Kong is 2 years old and we are using deprecated /apis end points which would prevent an urgent update to recent Kong if the need ever a-raised.

We recently had todo an RDS update of Kong so currently have knowledge of what Kong is and how it works.
Now seems like a good time to lift Kong over this particular speed bump.

## How to test

This process has been preformed in CODE and produced no down time.

Zero down time; no one outside of the CAPI team should be aware this happened.

Altering an arrangement which has been in place for many years.
Potential to block CAPI access or fail rate limits wide open.

### DIFF
--- a/scripts/migration-apis-to-services/code/migrate-plugins.py
+++ b/scripts/migration-apis-to-services/code/migrate-plugins.py
@@ -24,7 +24,7 @@ while url != "":
             resp, content = http.request("http://localhost:8001/services/internal/plugins",
                     "PUT", body=json.dumps(payload),
                     headers={'content-type':'application/json'})
-            print(request)
+            print(content)
 
     url = ""
     if 'next' in parsed:
--- a/scripts/migration-apis-to-services/code/migrate-plugins.py
+++ b/scripts/migration-apis-to-services/code/migrate-plugins.py
@@ -5,29 +5,30 @@ url = "http://localhost:8001/plugins?size=10"
 print(url)
 
 while url != "":
-        http = httplib2.Http()
-        content = http.request(url)[1]
-        print(content)
+    http = httplib2.Http()
+    content = http.request(url)[1]
 
-        parsed = json.loads(content)
+    parsed = json.loads(content)
 
-        data = parsed['data']
-        for plugin in data:
-                name = plugin['name']
-                if name == "rate-limiting":
-                        consumer_id = plugin['consumer_id']
-                        config = plugin['config']
-                        print(consumer_id)
+    data = parsed['data']
+    for plugin in data:
+        name = plugin['name']
+        if name == "rate-limiting":
+            consumer_id = plugin['consumer_id']
+            config = plugin['config']
+            print(consumer_id)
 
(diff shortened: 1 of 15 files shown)


scripts/migration-apis-to-services/code/migrate-plugins.py
```diff
@@ -0,0 +1,34 @@
...
+            config = plugin['config']
+            print(consumer_id)
+
+            payload = {'name': name, 'consumer_id': consumer_id, 'config': config}
+            print(json.dumps(payload))
+
+            resp, content = http.request("http://localhost:8001/services/internal/plugins",
+                    "PUT", body=json.dumps(payload),
+                    headers={'content-type':'application/json'})
+            print(request)
```
JustinPinner, line 27: Just wondering where `request` has come from in this print statement 🤔

tonytw1: Ah. 
Incorrectly cut and paste from instance to local git (bad but couldn't work out the ssm scp format within a reasonable time frame).
Corrected.

JustinPinner (commented):
So if I'm following this, we create scripts that can be run to configure the new endpoints, while leaving the deprecated `/apis` in place until we're ready to remove it? If that's the case I think it looks good. But I just saw that there's been another update, so this will be a comment until I've had a chance to look at what that was.

JustinPinner (approved):
As said before, this looks good to me. Will approve but, as always, other opinions are welcome 👍

tonytw1:
> So if I'm following this, we create scripts that can be run to configure the new endpoints, while leaving the deprecated `/apis` in place until we're ready to remove it? If that's the case I think it looks good. But I just saw that there's been another update, so this will be a comment until I've had a chance to look at what that was.

Yes.
In code we investigate is we are allowed to build the new service out in parallel while traffic continues to be server from the old /api. It looks like that it can.

PR #155, opened 2021-02-02
Update integration tests

tonytw1:
## What does this change?

Restores the local integration tests (sbt it:test).
Levels the containers we test against up to the same versions as live.

Migrates the container setup and teardown to docker-compose so that it works the same is local dev.

## How to test

```
sbt it:test
```


integration-test/scala/integration/fixtures/BonoboKeysTableFixture.scala
```diff
@@ -17,22 +17,32 @@ trait BonoboKeysTableFixture extends DynamoDbClientFixture with BeforeAndAfterAl
     val keySchema: List[KeySchemaElement] = List(
       new KeySchemaElement().withAttributeName("hashkey").withKeyType(KeyType.HASH),
       new KeySchemaElement().withAttributeName("rangekey").withKeyType(KeyType.RANGE))
-    val indexKeySchema: List[KeySchemaElement] = new KeySchemaElement().withAttributeName("keyValue").withKeyType(KeyType.HASH) :: Nil
```
tonytw1, line 20: Interesting way to construct a list... 
opinionated revert back to Seq(a) it required less vocabulary for a visiting polyglot developer to parse.

JustinPinner (approved):
This is not exactly familiar territory for me but I think what I see looks good. Will approve pending other opinions 👍

PR #156, opened 2021-02-04
Post migration cleanup; don't setup deprecated /apis locally.

tonytw1:
## What does this change?

Use plugins and routes helper paths under named service to simplify setup.


JustinPinner (approved):
LGTM 👍

PR #157, opened 2021-03-03
Deploy bonobo with a bionic Java 8 base image

tonytw1:
## What does this change?

Deploy with `ubuntu-bionic-capi` base image which has the same java8 and and logstash roles as the old xenial image.

Screenshot 2021-03-03 at 13 04 51

## How to test

Bonoho code still works.
Logstash still stashing
instance logstash logs not showing errors.


JustinPinner (approved):
👍

PR #159, opened 2021-06-18
Added extended at field to keys table

rowannekabalan:
## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

### DIFF
--- a/app/views/renderKeysTable.scala.html
+++ b/app/views/renderKeysTable.scala.html
@@ -12,6 +12,7 @@
                 <th>Tier</th>
                 <th>Status</th>
                 <th>Created at</th>
+                <th>Extended at</th>
             </tr>
         </thead>
         <tbody>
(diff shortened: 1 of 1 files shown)


app/views/renderKeysTable.scala.html
```diff
@@ -25,6 +26,7 @@
                     <td>@key.kongKey.tier.friendlyName</td>
                     <td>@key.kongKey.status</td>
                     <td>@key.kongKey.createdAt.toString("dd-MM-yyyy hh:mma")</td>
+                    <td>@key.bonoboUser.additionalInfo.extendedAt</td>
```
JustinPinner, line 29: This is a bit tricksy because the `extendedAt` value is an `Option[Long]` so we may have to do a little extra manipulation of that to translate it into a nice human readable datetime format.

PR #162, opened 2021-07-07
Rollback and reintroduce extended at field

rowannekabalan:
## What does this change?
This reverts commit 9200b1716b0befb322c58fd854a005fe4a42c33c (pushed directly to master by mistake) and reintroduces the changes to follow best practice.

## What was done?
A new field has been added to the All Keys view page in bonobo to display the key's extendedAt field if it exists.

## How to test
Run bonobo locally and point the application.conf file to the PROD dynamoDB tables (the extendedAt field does not exist in CODE).
Check All Keys view page for the extendedAt field or search for a specific user that you know has an extendedAt date and look for the date being displayed in correct format.
(You can also just look at Bonobo PROD as the change is already reflected there)

Field should display correct date in correct format if it exists and should handle null values gracefully by displaying nothing.

This is a low-risk change.

## Images
Example of what the field looks like:
Screenshot 2021-07-07 at 11 10 59

### DIFF
--- a/app/views/renderKeysTable.scala.html
+++ b/app/views/renderKeysTable.scala.html
@@ -1,8 +1,4 @@
-@import org.joda.time.DateTime
-
 @(keys: List[BonoboInfo], lastDirection: String, hasNext: Boolean)
-
-
 
 <div class="col-md-12">
     <table class="table table-responsive table-striped table-hover">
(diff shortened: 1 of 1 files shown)


JustinPinner:
Ha. This is a tricky one as it appears just to be removing blank lines - because the change is already in place due to that accidental push to master. I wonder if there's another way to reflect the changes without resorting to a full PR. Let's investigate?

JustinPinner (approved):
As discussed, we'll approve and merge this to capture the commentary of what happened, rather than revert and reapply in separate steps (the original changes merged prematurely were good anyway). The net effect to the current master branch is that a few empty lines will be removed, so nothing unexpected should come of that. 👍

PR #164, opened 2021-07-15
Updating Dynamo search function to find users with no keys

rowannekabalan:
## What does this change?
This allows searching for users that have no keys in Bonobo, and consequently adding keys to those users. 

Why was this done?
Bonobo search used to only bring up users who have keys, which meant if a users only key was deleted, there was no way to recreate their key in the UI. 

## How to test

Locally:
-Run Bonobo locally following the steps in the readme file, make sure the application is pointed to CODE DynamoDB tables (in the application.conf file)

On CODE:
-Go to bonobo-code.capu.gutools.co.uk

-Create a user and a key
-Delete that user's key
-Search for user using their name, email, or company name
-A record with that user's details should show up (and the key field should appear empty). You should be able to click on that user record and add/delete keys as you wish. If a user has no keys, the Created at field would display the user's creation date rather than the key's creation date

Better user experience when recreating keys that have been deleted.

This is a low risk change.


app/store/Dynamo.scala
```diff
@@ -245,6 +248,21 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
...
+        val createdAt = user.additionalInfo.createdAt
+        val emptyKey = new KongKey(bonoboId, "", "", 0, 0, Tier.Internal, "", createdAt, "", None, "")
+        val userRecord = users.find(_.bonoboId == bonoboId)
+        userRecord map {
+          BonoboInfo(emptyKey, _)
+        }
+      } else None
+    }
+  }
+
```
JustinPinner, line 265: I was wondering if this could be simplified a bit. For example, we re-run `users.find` but I think we already have the `users` we need. And that `None` looked strange sitting where it was. What do you think of something like this (I haven't tried it but IJ looks happy!);
```
  private def getUsersWithNoKeysInfo(users: List[BonoboUser]): List[BonoboInfo] = {
    val keys = getKeysForUsers(users)
    users.filterNot(u => keys.exists(_.bonoboId == u.bonoboId)).map { keylessUser =>
      val fakeKey = new KongKey(keylessUser.bonoboId, "", "", 0, 0, Tier.Developer, "", keylessUser.additionalInfo.createdAt, "", None, "")
      BonoboInfo(fakeKey, keylessUser)
    }
  }
```
Here we gather all the keys for all the users and then we use the `filterNot` to narrow down our users to those whose `bonoboId`s don't appear in the `keys` list. Then we map over any `keylessUser`s coming out of that, creating fake keys for the `BonoboInfo` return type.

What do you think?

rowannekabalan: Thanks Justin! I think this looks good, I  knew for a fact that there had to be a better way to write this but I went with the little Scala knowledge that I have :) I'm happy to test this change out and push it

JustinPinner: You're welcome 😄  Glad it worked too! 🥳

JustinPinner (approved):
I think there's a wider question here about absent keys as a valid condition, but to get things moving this looks great 👍

rowannekabalan:
> I think there's a wider question here about absent keys as a valid condition, but to get things moving this looks great 👍

Agreed @JustinPinner! This was the least disruptive way to get around the strict KongKey conditions everywhere, but I guess this could be revisited as part of our Bonobo revamp :)

PR #165, opened 2021-09-01
Rate limiting per second

rowannekabalan:
## What does this change?
Adds support for rate limiting per second in Bonobo (which Kong already supports, but we weren't using). 

## How to test
TBC

TBC

TBC


app/models/models.scala
```diff
@@ -143,7 +144,7 @@ object RegistrationType {
   }
 }
 
-case object DeveloperRegistration extends RegistrationType {
+case object DeveloperRegistration extends RegistrationType {:
```
Fweddi, line 147: this might be a breaking typo?

PR #166, opened 2021-09-16
Add key extension logic

rowannekabalan:
## What does this change?
This moves the developer key extension/deletion logic from the User level to the Key level. This was done to avoid deleting a user and all their keys. Once a user doesn't have any keys left they will be deleted automatically.

## How to test
To be end-to-end tested on CODE

Users should be able to extend/delete their developer keys.

This is a major change to the way we handle key cleanup logic, so we need to make sure we didn't break anything.


app/store/Dynamo.scala
```diff
@@ -179,6 +181,14 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
       new AttributeUpdate("requests_per_minute").put(kongKey.requestsPerMinute),
       new AttributeUpdate("status").put(kongKey.status),
       new AttributeUpdate("tier").put(kongKey.tier.toString),
+      kongKey.remindedAt match {
+        case Some(date) => new AttributeUpdate("extendedAt").put(date)
```
alinaboghiu, line 185:
```suggestion
        case Some(date) => new AttributeUpdate("remindedAt").put(date)
```

app/store/Dynamo.scala
```diff
@@ -179,6 +181,14 @@ class Dynamo(db: DynamoDB, usersTable: String, keysTable: String, labelTable: St
...
+        case None => new AttributeUpdate("extendedAt").delete()
```
alinaboghiu, line 186:
```suggestion
        case None => new AttributeUpdate("remindedAt").delete()
```

app/store/Dynamo.scala
```diff
@@ -416,6 +429,9 @@ object Dynamo {
       .withList("labelIds", labelIds.asJava)
 
     kongKey.productUrl.foreach(productUrl => item.withString("productUrl", productUrl))
+    kongKey.remindedAt.foreach(remindedAt => item.withLong("remindedAt", remindedAt))
+    kongKey.extendedAt.foreach(remindedAt => item.withLong("remindedAt", remindedAt))
```
alinaboghiu, line 433:
```suggestion
    kongKey.extendedAt.foreach(extendedAt => item.withLong("extendedAt", remindedAt))
```

app/logic/DeveloperFormLogic.scala
```diff
@@ -43,26 +43,30 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) {
...
+
+  def deleteKeyByUser(key: KongKey): Future[_] = {
+    Future {
+      for {
+        _ <- kong.deleteKey(key.kongConsumerId).recover {
+          case Kong.GenericFailure(txt) if txt.contains("404") => ()
+        }
+      } yield {
+        dynamo.deleteKey(key)
+        if (dynamo.getKeysWithUserId(key.bonoboId) == 0) {
```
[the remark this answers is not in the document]
rowannekabalan: Ah good catch thank you!

alinaboghiu (approved):

PR #167, opened 2021-09-22
Rate limit per second

rowannekabalan:
## What does this change?
Adds rate limiting per second to Bonobo (which is already supported by Kong but we weren't using). 

## How to test
Can be tested locally or on CODE by running the app and attempting to edit existing keys or create new ones.

 This is for us to experiment with and see whether it adds any value in handling the spikes of 500s that we've been seeing in concierge lately.

This is a relatively low risk change as the rate limit per second field is optional to make sure existing keys keep working as they should.


Fweddi (approved):
Requests per seconds optional for backwards compatibility. LGTM - have we tested on CODE?

rowannekabalan:
> Requests per seconds optional for backwards compatibility. LGTM - have we tested on CODE?

Thanks! Yes this is the branch currently deployed on CODE, so you can play around with updating the rate limits for your existing CODE keys.

PR #185, opened 2022-05-19
Handle other options in developer form

rowannekabalan:
## What does this change?

This ensures that if I user selects the "Other" option in the "Reason for key" or "Company sector" fields, an additional text field would appear for them to give us more information.

Changes made:

- Adding the custom fields everywhere they need to be
- Handling validation and optionality of those custom fields
- Adding the custom fields to the User and Key models 
- Updating all CRUD operations to include the new fields
- Ensuring meaningful error messages show up in forms
- Scripts to handle showing/hiding/clearing custom fields

## How to test

Tested locally. To be tested on CODE.

Nothing breaks and we successfully capture new data about how people use our keys.

Given the scale of the changes it is important to ensure nothing broke and the key/user creation/deletion/extension processes work as expected.

## Accessibility

- [ ] [Tested with screen reader](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#screen-readers)
- [ ] [Navigable with keyboard](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#keyboard-navigation)
- [ ] [Colour contrast passed](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#colour-contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#use-of-colour)


app/logic/ApplicationLogic.scala
```diff
@@ -181,8 +181,8 @@ class ApplicationLogic(dynamo: DB, kong: Kong) extends Logging {
     case None => f
   }
 
-  private def saveKeyOnDB(userId: String, consumer: ConsumerCreationResult, rateLimits: RateLimits, tier: Tier, reasonForKey: Option[String], productName: String, productUrl: Option[String], labelIds: List[String]): Unit = {
-    val newKongKey = KongKey(userId, consumer, rateLimits, tier, reasonForKey, productName, productUrl)
+  private def saveKeyOnDB(userId: String, consumer: ConsumerCreationResult, rateLimits: RateLimits, tier: Tier, reasonForKey: Option[String], customReasonForKey: Option[String], productName: String, productUrl: Option[String], labelIds: List[String]): Unit = {
+    val newKongKey = KongKey(userId, consumer, rateLimits, tier, reasonForKey, customReasonForKey, productName, productUrl)
```
JustinPinner, line 185: Just a question: does putting new data in a spot that was previously occupied by something else (in this case `customReasonForKey` sits where `productName` used to, and so `productName` occupies `productUrl`'s former location) cause inconsistencies in the dynamo data structure? I assume it'd go in just fine as it's stored as a json object but can we get consistent from old records when querying it later? This may turn out to be a pointless question when I've thought about it some more!

JustinPinner (approved):
This looks good to me 👍

app/logic/ApplicationLogic.scala
JustinPinner: As suspected - it's all good as the order in the objects doesn't matter at the time the data is written

PR #186, opened 2022-05-26
Make "reasonForKey" optional on the server side

rowannekabalan:
## What does this change?

This makes the "reasonForKey" attribute optional on the server-side and keeps in mandatory on the client-side. That is to avoid errors in old keys that do not have this attribute and to avoid added complexity in the code. This came out of an incident reported by the Journalism team where key extensions were failing due to the "reasonForKey" attribute not existing.

## How to test

Testing locally and on CODE. 

Key extensions should work fine.

## Accessibility

- [ ] [Tested with screen reader](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#screen-readers)
- [ ] [Navigable with keyboard](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#keyboard-navigation)
- [ ] [Colour contrast passed](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#colour-contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://guardian.github.io/source/?path=/docs/docs-06-accessibility--page#use-of-colour)


JustinPinner (approved):
Looks good to me 👍

PR #187, opened 2022-07-25
Extract script

JustinPinner:
## What does this change?

We've been working with version(s) of this script for a while - it's pretty basic but gets what we need out, and as it doesn't expose any secrets, we may as well bring it into this repo I think.

## How to test

Check it out, get credentials from Janus, set `const STAGE = codeStage`  and run it with `node extract.js`. You'll end up with a `CODE-users.csv`.

If the data in the csv file looks right, SUCCESS!

Repeatedly running it on PROD with very large date ranges might be a nuisance, so test any changes on CODE first. Typically it's only used to run to cover the last few months and doesn't appear to cause any overhead for Dynamo.

## Images

N/A

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)
-   [x] This is a command-line script.


rowannekabalan (approved):
This is a very handy script to have! LGTM 👍

PR #189, opened 2022-08-11
Show last used times for keys

fredex42:
## What does this change?

- Updates Bonobo to display "days since this key was last used" data that will be input to the `keys` table by KongKeyMon

**NOTE** Requires https://github.com/guardian/kongkeymon/pull/12 to push data into the table. However, it should be able to run on non-updated tables without a problem, simply displaying "Not Detected" for every key

Screenshot 2022-08-22 at 17 19 34

## How to test

- View the keys table and find one of yours that has not been used
- Click it to get to the user and notice the same values in the 'Keys for this user' table
- Make a couple of requests via Open Platform code
- Give it a couple of seconds for the data to come through
- Run the KongKeyMon process that extracts the data (see seperate forthcoming PR)
- Refresh Bonobo and see the new data

Easy viewing of active/inactive keys

n/a


app/models/models.scala
```diff
@@ -73,7 +75,18 @@ case class KongKey(
...
-  rangeKey: String)
+  rangeKey: String,
+  lastUsedTimestamp: Option[Long]) {
+
+  def lastUsedDaysAgo = lastUsedTimestamp match {
+    case None => "Not detected"
+    case Some(ts) =>
+      val nowTime = Instant.now()
+      val lastUsedTime = Instant.ofEpochMilli(ts)
+      val timePeriodDays = (nowTime.getEpochSecond - lastUsedTime.getEpochSecond) / 86400
```
paulmr, line 86: How about:

```scala
val timePeriodDays = lastUsedTime.until(nowTime, DAYS)
```

(after importing `java.time.temporal.ChronoUnit.DAYS` of course).

paulmr: or maybe:

```scala
val timePeriod = Duration.between(lastUsedTime, nowTime)
```

then you can be a bit more flexibly about how you stringify it (although not as much as a calender-based option like `Period`).

app/models/models.scala
```diff
@@ -73,7 +75,18 @@ case class KongKey(
   customReasonForKey: Option[String],
   productName: String,
   productUrl: Option[String],
-  rangeKey: String)
+  rangeKey: String,
+  lastUsedTimestamp: Option[Long]) {
```
paulmr, line 79: I still feel like we could save this as a duration? I am just imagining myself asking the perennial question, "hmm, is this timestamp in seconds or milliseconds?"

Could we at least perhaps name it something like `lastUsedEpochSeconds`?

fredex42: @rowannekabalan would be good to get your input here on compatibility with Gibbons

rowannekabalan (approved):
Looks good to me 👍 However I have a question, is there any specific reason why we need the `safeGetLong` function for this field but not for other `Long` type fields? @fredex42

fredex42:
@rowannekabalan I could switch the others over to `safeGetLong` as well. The reason I put it in was that I was consistently getting `null` values returned _within_ the Option, so I explicitly caught that problem.  This might be to do with the column itself being absent in some of the records?

PR #192, opened 2022-08-31
Feature / count keys

fredex42:
## What does this change?

Shows stats on the front page about how many keys are active (i.e. unique keys used within a given time period)
This is calculated daily by KongKeyMon (https://github.com/guardian/kongkeymon/pull/14) and output into DynamoDB from where Bonobo picks it up

## How to test

- Run the KongKeyMon job as detailed on the above linked PR
- Deploy this to CODE and then load the frontpage. You should see a new box at the top of the screen (see image)

Able to see key usage stats

n/a

## Images

Screenshot 2022-08-31 at 16 53 24


macros/src/main/scala/models/DynamoMacros.scala
```diff
@@ -0,0 +1,30 @@
...
+import com.amazonaws.services.dynamodbv2.document.Item
+
+import scala.language.experimental.macros
+import reflect.runtime.universe._
+import scala.reflect.macros.blackbox
+
+object DynamoMacros {
+  def getField[T](record: Item, fieldName: String): Option[T] = macro getFieldImpl[T]
+
+  def getFieldImpl[T: c.WeakTypeTag](c: blackbox.Context)(record: c.Tree, fieldName: c.Tree): c.Tree = {
```
JustinPinner, line 12: I'm finding this kind of hard to justify. If you can find any other examples of this approach used elsewhere in any of our repos, then we can potentially take that as a precedent, but my current feeling is that, whilst this may work, we do know what we're asking for when we're reading in the record, and so ought to be able to deal with it in a more straightforward and easier to read way.

fredex42: 👍 Fair enough, done.

JustinPinner: Thanks. I think this looks good to go now 👍

JustinPinner (approved):
LGTM 👍

PR #193, opened 2022-09-07
Version of Kong that runs on m1

fredex42:
## What does this change?

- Updates docker-compose to run Kong v1.5 via amd64 emulation that is present in Docker desktop
- v1.5 is the minimum version that has bidirectional compatibility for version 0.14 and recent ones
- Some parameters have been renamed, the code is updated to reflect the new names
- Also adds a persistent storage volume in docker-compose, so you won't lose all changes in Kong if you shut the container down

But I _want_ to reset my Kong database! - No problem, just run `docker volume rm scripts_pgdata` with the stack shut down.

## How to test

- Follow the standard instructions to set it up; everything should work

To verify compatibility on updates:
- take a pgsql dump of a 0.14 database (for example, the CODE environment)
- use docker-compose to only start the database pod
- use `docker cp` to copy your sql dump into the database container, then `docker exec` to shell into it and run `pg_restore` to set your local database up as a mirror
- now start up kong and follow the logs with `docker log -f`
- you should see it tell you that you need to update.  
- shell into the Kong container with `docker exec` and run the recommended update command
- everything should then work

- Able to run Kong locally on M1 (apple silicon) Macs
- Ready to perform an upgrade to Kong 1.5 in our real environments

- Don't merge this until we have updated Kong in the CODE and PROD environments, because of the change to parameter names. The code is not backwards compatible to v0.14


rowannekabalan (approved):
Ran perfectly on my M1 🎉

PR #194, opened 2022-09-07
Update README to indicate how to run locally via CODE Kong

fredex42:
## What does this change?

No code updates. Just updates to the README

## How to test

Read and see if makes sense. Then try what is recommended!

We're able to test with Kong on an M1 Mac before the upgrade work is completed

n/a


JustinPinner (approved):
Nice 👍

PR #188, opened 2022-08-01
Add email verification

rowannekabalan:
## What does this change?

This updates the developer sign up flow to include an email verification step. Details can be found [here.](https://docs.google.com/document/d/1EThtYQigLa24xoaE4Niec-3iJs5xpDxLDwT5OzjhSak/edit)

## How to test

Deploy to CODE and enable emails in config - you should be able to test the whole sign up flow end-to-end.

## How can we measure success?

People with fake email addresses can't obtain API keys any more.

## Have we considered potential risks?

This is a medium risk change. It introduces potentially breaking changes to the sign up logic, and includes storage changes, UI changes, and more. It has been tested end-to-end but we should keep and eye on it after it goes out.

## Images

In document attached above.

PR #196, opened 2022-12-07
Add endpoint to delete unverified keys

rowannekabalan:
## What does this change?

This adds an endpoint to delete an unverified user and their key. This is related to this [PR](https://github.com/guardian/bonobo/pull/188) which introduced email address verification. This meant we are ending up with unverified users and keys which need cleaning up. That will done via a new lambda in [Gibbons](https://github.com/guardian/gibbons/pull/42).

## How to test

This has been tested by running the Gibbons lambda in CODE. You could also run the lambda in DryRun mode in PROD (#To-Do).

Unverified users and keys should be regularly deleted from Kong and Dynamo.

This is a low-risk change.


app/controllers/DeveloperForm.scala
```diff
@@ -114,6 +114,23 @@ class DeveloperForm(override val controllerComponents: ControllerComponents, dyn
     }
   }
 
+  def deleteUnverifiedUserAndKey(userId: String) = Action.async { implicit request =>
+    Future {
+      dynamo.getUserWithId(userId)
+    } flatMap { user =>
+      if (user.exists(_.additionalInfo.verificationSentAt.nonEmpty)) {
+        logic.deleteKeyFromDynamoAndPossiblyUser(dynamo.getKeysWithUserId(userId).head).recover {
```
JustinPinner, line 122: Is `.head` intentional here or do we expect a 1:1 relationship between the user and keys? I can't recall if it's possible to register more than one key from the outside world.

app/controllers/DeveloperForm.scala
```diff
@@ -114,6 +114,23 @@ class DeveloperForm(override val controllerComponents: ControllerComponents, dyn
...
+          case err =>
+            awsEmail.sendEmailDeletionFailed(user.get.bonoboId, err)
```
JustinPinner, line 124: If the user/key is unverified, should we bother sending another email to the same address?

app/logic/DeveloperFormLogic.scala
```diff
@@ -54,21 +54,28 @@ class DeveloperFormLogic(dynamo: DB, kong: Kong) extends Logging {
...
       } yield {
-        dynamo.deleteKey(key)
-        if (dynamo.getKeysWithUserId(key.bonoboId).isEmpty) {
-          for {
-            _ <- kong.deleteConsumer(key.kongConsumerId).recover {
-              case Kong.GenericFailure(txt) if txt.contains("404") => ()
-            }
-          } yield dynamo.deleteUser(dynamo.getUserWithId(key.bonoboId).get)
-        }
+        deleteKeyFromDynamoAndPossiblyUser(key)
```
JustinPinner, line 64: Just a sense check here: are we at risk of deleting the dynamo record(s) if the kong request fails with anything other than a 404?

app/controllers/DeveloperForm.scala
rowannekabalan: It's indeed not possible to register more than one key to the same owner from the outside world so we do expect that it's 1:1 in this scenario but the use of `.head` is just for the off chance that someone has manually manipulated the keys (it wouldn't cause any harm as we never delete a user if they still own any keys)

app/logic/DeveloperFormLogic.scala
rowannekabalan: If I'm not misinterpreting this is saying if Kong fails to delete the key, then only recover from that failure if it's a 404 (in that case continue with deleting from Dynamo). Otherwise the Kong failure would just break the for comprehension and not continue the next steps.

app/controllers/DeveloperForm.scala
rowannekabalan: Also on the off chance that they own more than one key, the order of them doesn't matter as it's an unverified user and should be deleted anyway, so their keys will get deleted one by one with each run of the lambda until their last key is gone and the user gets deleted as well.

JustinPinner (approved):
Sorry - meant to 👍 this yesterday!

PR #197, opened 2022-12-29
Reduce developer key rate limits

rowannekabalan:
## What does this change?

Reduces developer key rate limits.

## How to test

Tested in CODE
Screenshot 2022-12-29 at 11 05 09

New keys should be created with the new limits. Existing keys should still have their previous limits.


JustinPinner (approved):
👍

PR #198, opened 2023-01-06
Disable register button

rowannekabalan:
## What does this change?

Disables Register button in developer registration form once reCAPTCHA is displayed to avoid multiple submits, as completing the reCAPTCHA challenge auto-submits the form.

This is to remediate exceptions we've been seeing in Bonobo which we think are due to people submitting the form twice.
Screenshot 2023-01-06 at 13 27 22

## How to test

Tested in CODE.

The number of "no reCAPTCHA token found" exceptions should drop.

No risk change.

## Images

Screenshot 2023-01-06 at 13 23 07


JustinPinner (approved):
👍

PR #199, opened 2023-01-11
Log headers for failing recaptcha requests

rowannekabalan:
## What does this change?

We're seeing reCAPTCHA errors in the logs and we're trying to understand where they're coming from.
Screenshot 2023-01-11 at 07 58 14


fredex42 (approved):
👍 let's hope it tells us some useful info!

PR #203, opened 2023-03-22
Remove defunct script and config

rowannekabalan:
## What does this change?

This removes an old script that finds which API keys currently being used are issued under Developer tier. 
This script was used before we had KongKeyMon for monitoring key usage in order to debug missing developer keys that were mistakenly deleted by Gibbons. There is no further need for this functionality so we are deleting it.

## How to test

Nothing should be affected by this change.


JustinPinner (approved):
🧹 👍

PR #204, opened 2023-03-23
Graviton compatibility for Bonobo

fredex42:
## What does this change?

- Updates base image to in-support Ubuntu
- Moves to an ARM base image
- Remove logstash and set up devx's fluentbit instead because logstash kinesis forwarding does not work on ARM

## How to test

Deploy it and check that it works. Currently deployed to CODE.

Lower costs and (slightly) happier management and DevX

If this fails it's expected to fail fast, as in not start up. If it's running then it will probably carry on running.


JustinPinner (approved):
LGTM 👍

